PR #96, opened 2024-08-28
[#79] HandleTrie wrapper and benchmark

### LINKED ISSUE #79: PoC - Create a wrapper to HandleTrie in python
EPIC

* 

HandleTrie is a class implemented in C++ here: https://github.com/singnet/das-attention-broker/blob/master/src/attention_broker_server/HandleTrie.h

It's basically a hashmap to map from atom's handle (MD5 hash string) to a generic object.

We want to build a wrapper to this class in python in such a way that when the python class is instantiated, it uses the C++ implementation. In this case, the hashmap will map from handle -> Dict[str, any]

The result of the PoC is a report comparing performance of:

* (a) A C++ program adding and making queries to HandleTrie
* (b) A Python program adding and making queries to the python wrapper of HandleTrie
* (c) A Python program adding and making queries to python's dict

eddiebrissow:
closes #79


angeloprobst (approved):

marcocapozzoli (approved):

experiments/handletrie_benchmark/README.md
```diff
@@ -0,0 +1,69 @@
...
+|---|---|---|---|---|---|
+|C++ Handle Trie      |0.00129|0.15719|1.86882|23.91808|149.15487*|
+|C++ unordered map std|0.00054|0.15927|2.12357|22.52662|-|
+|C++ map std          |0.00151|0.32405|5.21409|71.53995|561.42147*|
+|Python's Dict        |0.00037|0.06631|0.89053|11.75517|79.204370*|
+|CPython Handle Trie  |0.00130|0.38828|5.00333|63.05321|510.58422*| 
+|NanoBind Handle Trie |0.00281|0.52542|6.52422|84.37761|651.15217*| 
+|PyBind Handle Trie   |0.01159|1.41426|15.5277|168.3574|1159,28535*|
+
+\* The 60,000,000 column values may not be accurate, as the test fill up all the available memory on the test machine, leading to inaccurate results.
```
andre-senna: I'd add another comment here explaining that comparison between C++ and Python objects is somehow affected by the fact that the operations being measured in each one are not exactly the same. For instance, Time in C++ implementation is taking into account the random number generation used to populate keys.

andre-senna (approved):

Issue #166, opened 2024-12-23
Link creation agent C++


eddiebrissow:
- DAS Node server: receive and process messages
- Link search 
- Link creation algorithm
- Tests

Links to Epic #128

PR #195, opened 2025-01-22
[das#166] Link Creation Agent DAS Node communication

eddiebrissow:
links to #166


src/cpp/link_creation_agent/BUILD
```diff
@@ -0,0 +1,11 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "link_creation_agent_lib",
+    srcs = glob([ "src/das_node/*.cc"]),
+    hdrs = glob(["include/*.h", "include/das_node/*.h", "include/das_node/das_server/*.h", "include/das_node/link_creation/*.h", "include/das_node/query/*.h", "include/utils/*.h"]),
```
angeloprobst:
```suggestion
    hdrs = glob(["include/**/*.h"]),
```

eddiebrissow: done

angeloprobst (approved):

src/cpp/link_creation_agent/agent.cc
```diff
@@ -0,0 +1,83 @@
...
+    link_creation_node_server->graceful_shutdown();
+    if (agent_thread != NULL && agent_thread->joinable()) {
+        agent_thread->join();
+        agent_thread = NULL;
+    }
+}
+
+void LinkCreationAgent::run() {
+    while (true) {
+        this_thread::sleep_for(chrono::seconds(loop_interval));
```
andre-senna, line 36: The thread is supposed to sleep only when there's no further requests to process. Something like:

```
while true
    if (request queue is empty)
        then sleep
        else process request
```

src/cpp/link_creation_agent/agent.h
```diff
@@ -87,10 +91,12 @@ namespace link_creation_agent
 
 
         // Other attributes
-        LinkCreationService service;
-        vector<LinkCreationAgentRequest> request_buffer;
-        query_node::QueryNode query_node_client;
-        lca::LinkCreationNode link_creation_server;
-        das::ServerNode das_client;
+        LinkCreationService *service;
+        vector<LinkCreationAgentRequest> *request_buffer;
```
andre-senna, line 95: vector is not thread safe. Shouldn't we use a thread safe queue here?

andre-senna (approved):

src/cpp/link_creation_agent/agent.h
eddiebrissow: only one thread is accessing the vector for now

PR #218, opened 2025-01-28
[#das166] Create Link Agent main features

eddiebrissow:
links to #166

### DIFF
--- a/src/cpp/link_creation_agent/agent.cc
+++ b/src/cpp/link_creation_agent/agent.cc
@@ -1,8 +1,9 @@
 #include "agent.h"
-#include "RemoteIterator.h"
 
 #include <fstream>
 #include <sstream>
+
+#include "RemoteIterator.h"
 
 using namespace std;
 using namespace link_creation_agent;
@@ -14,6 +15,8 @@ LinkCreationAgent::LinkCreationAgent(string config_path) {
     load_config();
     link_creation_node_server = new LinkCreationNode(link_creation_server_id);
     query_node_client = new DASNode(query_node_client_id, query_node_server_id);
+    service = new LinkCreationService(thread_count);
+    das_client = new ServerNode(das_client_id, query_node_server_id);
 
     this->agent_thread = new thread(&LinkCreationAgent::run, this);
 }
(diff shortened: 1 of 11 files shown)


src/cpp/link_creation_agent/agent.cc
```diff
@@ -14,38 +15,91 @@ LinkCreationAgent::LinkCreationAgent(string config_path) {
     load_config();
     link_creation_node_server = new LinkCreationNode(link_creation_server_id);
     query_node_client = new DASNode(query_node_client_id, query_node_server_id);
+    service = new LinkCreationService(thread_count);
+    das_client = new ServerNode(das_client_id, query_node_server_id);
```
andre-senna, line 19: Having an object named `das_client` of a class named `ServerNode` is bizarre.

If I understood it right, `das_client` the object used by this agent to issue query requests to a Query Agent. Is this correct? If yes, it is, indeed, a client. SO you should rename this `ServerNode` class.

src/cpp/link_creation_agent/servive.cc
```diff
@@ -0,0 +1,36 @@
...
+    thread_pool.enqueue(job);
+}
+
+void LinkCreationService::create_link(Link &link, ServerNode &das_client)
+{
+    std::unique_lock<std::mutex> lock(m_mutex);
+    cout << "Creating link" << endl;
+    vector<string> link_tokens = link.tokenize();
+    das_client.create_link(link_tokens);
+    m_cond.notify_one();
```
andre-senna, line 35: Please doble check if this is correct. I didn't see where someone is being put to sleep on this cond.
(I'm not saying it's wrong, this is just a heads up for you to double check whether it's correct)

andre-senna (approved):
Approved to merge but you should consider using `this->` prefixing all uses of class variables.

src/cpp/link_creation_agent/agent.cc
eddiebrissow: I'll change this in the next PR

src/cpp/link_creation_agent/servive.cc
eddiebrissow: This method is called inside the thread pool, this is to prevent concurrency while using the das server client. I'll double check in the next PR

PR #219, opened 2025-01-28
[das#166] Link Creation Docs and Tests

### LINKED ISSUE #166: Link creation agent C++

eddiebrissow:
links to #166

### DIFF
--- a/src/cpp/link_creation_agent/README.md
+++ b/src/cpp/link_creation_agent/README.md
@@ -5,9 +5,22 @@
 
 DAS Link Creation Agent (DAS LCA), process a query and create links using the result of the query and a custom template. The service can execute n times a request to update the database with relevant links over the time.
 
-![alt](doc/assets/das_link_creation_hla.png)
+### Request
 
-Query Example
+The request must have 6 elements:
+1. Query (string): A valid Query Agent query (see Query Example)
+2. Link Create Template (string): A valid link create template (see Link Create Template Example)
+3. Max number of query response (int)
+4. Repeat (int): 0 to run once, -1 to run infinitely, 1 or higher to run this number of times.
+5. Context (string): Query context for Attention Broker requests 
+6. Update Attention Broker flag (bool): true or false to update or not the Attention Broker
+
+Example:
+```
+LINK_TEMPLATE", "Expression", "3", "NODE", "Symbol", "Similarity","VARIABLE", "V1", "VARIABLE", "V2", "LINK_CREATE", "Similarity", "2", "1", "VARIABLE", "V1", "VARIABLE", "V2", "CUSTOM_FIELD", "truth_value", "2", "CUSTOM_FIELD", "mean", "2", "count", "10", "avg", "0.9", "confidence", "0.9", "10", "0", "test", "false"
+```
+
+#### Query Example
 
 ```
 LINK_TEMPLATE Expression 3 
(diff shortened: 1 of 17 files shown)


src/cpp/link_creation_agent/README.md
```diff
@@ -5,9 +5,22 @@
 
 DAS Link Creation Agent (DAS LCA), process a query and create links using the result of the query and a custom template. The service can execute n times a request to update the database with relevant links over the time.
 
-![alt](doc/assets/das_link_creation_hla.png)
+### Request
 
-Query Example
+The request must have 6 elements:
```
andre-senna, line 10: Arguments 1. and 2. should be moved to the end instead of being the first ones. Otherwise you need to actually parse 1. and 2. before knowing the values of 3., 4., 5. and 6.

src/cpp/link_creation_agent/agent.cc
```diff
@@ -44,56 +43,54 @@ void LinkCreationAgent::run() {
...
+                is_from_buffer = true;
                 current_buffer_position = current_buffer_position % request_buffer.size();
                 lca_request = &request_buffer[current_buffer_position];
                 current_buffer_position++;
             }
         }
 
-        if (lca_request == NULL) {
+        if (lca_request == NULL ||
+            lca_request->last_execution + lca_request->current_interval > time(0)) {
```
andre-senna, line 71: What does this check do? I didn't understand what condition is being checked here.

src/cpp/tests/link_creation_agent_test.cc
```diff
@@ -0,0 +1,51 @@
...
+using namespace link_creation_agent;
+
+class LinkCreationAgentTest : public ::testing::Test {
+protected:
+    LinkCreationAgent* agent;
+
+    void SetUp() override {
+        // Create a temporary config file for testing
+        ofstream config_file("test_config.cfg");
+        config_file << "default_interval=1\n";
```
andre-senna: Is this 1 second or 1 millisecond? Or something else?

Anyway I have a couple of considerations regarding this "interval" mechanics.

* 1 second is too much and 1 millisecond is too few. I suggest 100 milliseconds of sleeping time.
* I don't think we should increase the duration of sleeping time as you did. I don't actually see the point of doing it. We definitely need the sleep but the I don't see why we would want to make the thread sleep for more than 100ms. Actually, allowing this sleeping time increase, as you did, may increase latency to answer requests unnecessarily.
* A minor point but... "interval" is a misleading name. The name of this variable should make more explicit that it's a duration of a sleeping time.

andre-senna (changes requested):

src/cpp/tests/link_creation_agent_test.cc
eddiebrissow: This is the request execution default interval in seconds, I'll rename it since the name is misleading.
The request that repeats will execute at arrive time, then 1 second later, then 2 seconds later and so on.

src/cpp/link_creation_agent/agent.cc
eddiebrissow: This checks if the request is empty or doesn't need to be processed yet

src/cpp/link_creation_agent/README.md
eddiebrissow: The order shouldn't matter, all arguments are required.

src/cpp/tests/link_creation_agent_test.cc
andre-senna: As I said, I suggest using a fixed interval of 100 milliseconds.

andre-senna (approved):

Issue #209, opened 2025-01-24
Implement link creation algorithm


andre-senna:
Implement link creation algorithm

Example of query and link creation template:

```
AND
    LINK_TEMPLATE Expression 3 
        NODE Symbol EVALUATION 
        LINK_TEMPLATE Expression 2
            NODE Symbol PREDICATE
            VARIABLE V1
        LINK_TEMPLATE Expression 2
            NODE Symbol CONCEPT
            VARIABLE V2
    LINK_TEMPLATE Expression 3 
        NODE Symbol EVALUATION 
        LINK_TEMPLATE Expression 2
            NODE Symbol PREDICATE
            VARIABLE V3
        LINK_TEMPLATE Expression 2
            NODE Symbol CONCEPT
            VARIABLE V4
```

```
LINK_CREATE Expression 3 1
    NODE Symbol IMPLICATION
    LINK_CREATE Expression 3 0
        NODE Symbol EVALUATION
        V1
        V4
    LINK_CREATE Expression 3 0
        NODE Symbol EVALUATION
        V3
        V2
    CUSTOM_FIELD truth_value 2
        CUSTOM_FIELD mean 2
            count 1
            avg 0.9
        confidence 0.9
```

Algorithm

```
Inputs:

(a) Query: object describing a pattern matching query
(b) Link creation template: object describing the link structure which is supposed to be created
(c) Identifier of a processor which is supposed to be used to create the links based on 
    QueryAnswer objects
(d) N: number of QueryAnswer objects which are supposed to be used to create links

Actually, (b) and (c) are mutually exclusive so the caller should pass one or 
the other but not both.

Algorithm:

(1) Submit (a) to the Query Agent, eventually passing (d)
(2) Iterate through QueryAnswer objects:
    If (b) is defined
        use QueryAnswer assignments to get the values for each variable and add the request 
        to create new links to a BUNDLE
    else (meaning that (c) is defined)
        pass the QueryAnswer object to the proper function and get the links to be built as 
        response, adding them to the BUNDLE
(3) After finishing iteration (either because the Iterator is depleted or because the limit of (d) have 
    been reached), request the addition of links in the BUNDLE.
```

IMPORTANT:

We should prioritize the implementation of step (2) using (c). Only AFTER this use case is properly implemented we are supposed to go for (b).

PR #236, opened 2025-01-30
[#209] Link Create Template parser

eddiebrissow:
links to #209

### DIFF
--- a/src/cpp/link_creation_agent/link_create_template.cc
+++ b/src/cpp/link_creation_agent/link_create_template.cc
@@ -2,42 +2,60 @@
 
 #include <iostream>
 #include <stdexcept>
+#include <algorithm>
 
 using namespace link_creation_agent;
 
-CustomField::CustomField(std::vector<std::string> args) {
-    if (args[0] != "CUSTOM_FIELD")
+bool is_number(const std::string& s) {
+    return !s.empty() &&
+           std::find_if(s.begin(), s.end(), [](unsigned char c) { return !std::isdigit(c); }) == s.end();
+}
+
+int string_to_int(const std::string& s) {
+    if (!is_number(s)) {
+        throw std::invalid_argument("Can not convert string to int: Invalid arguments");
+    }
+    return std::stoi(s);
+}
+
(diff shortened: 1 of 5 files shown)


src/cpp/link_creation_agent/link_create_template.cc
```diff
@@ -0,0 +1,243 @@
+#include "link_create_template.h"
+
+#include <algorithm>
+#include <iostream>
+#include <stdexcept>
+
+using namespace link_creation_agent;
+
+bool is_number(const std::string& s) {
```
andre-senna: Util functions outside classes should be declared `static` to prevent name clashing during linking.

```
static  bool is_number(const std::string& s) {
```

You can use them the same way you use non-static functions. The `static` keyword here means only that the function symbol is not supposed to be exposed outside this `.o` when linking to build an executable.

If you have purposely created a function outside classes which is supposed to be used in other files, you are supposed to prefix the function name with a namespace as you do with class names and put the function signature in the `.h`.

src/cpp/link_creation_agent/link_create_template.cc
```diff
@@ -0,0 +1,243 @@
...
+    return !s.empty() &&
+           std::find_if(s.begin(), s.end(), [](unsigned char c) { return !std::isdigit(c); }) == s.end();
+}
+
+int string_to_int(const std::string& s) {
```
andre-senna: (recurrent comment) Review ALL functions outside classes in ALL files.

src/cpp/link_creation_agent/link_create_template.cc
```diff
@@ -0,0 +1,243 @@
...
+    if (!is_number(s)) {
+        throw std::invalid_argument("Can not convert string to int: Invalid arguments");
+    }
+    return std::stoi(s);
+}
+
+std::string get_token(std::vector<std::string>& link_template, int cursor) {
```
andre-senna: (recurrent comment) Review ALL functions outside classes in ALL files.

andre-senna (approved):

PR #242, opened 2025-02-03
[#175] Client program that spawn other clients

marcocapozzoli:


src/cpp/client_spawner/SentinelServerNode.cc
```diff
@@ -0,0 +1,43 @@
...
+
+shared_ptr<Message> SentinelServerNode::message_factory(string &command, vector<string> &args) {
+    shared_ptr<Message> message = StarNode::message_factory(command, args);
+    if (message) {
+        return message;
+    }
+    if (command == WORKER_NOTIFICATION) {
+        return shared_ptr<Message>(new WorkerMessage(args[0], args[1]));
+    }
+    return shared_ptr<Message>{};
```
angeloprobst: This is not a change request, it's more like a tip.
The following code does the same, but it's a bit simpler and complies to the _Modern C++_ style recommendations.

```cpp
    auto message = StarNode::message_factory(command, args);  // <== here
    if (message) {
        return message;
    }
    if (command == WORKER_NOTIFICATION) {
        return make_shared<WorkerMessage>(args[0], args[1]);  // <== here
    }
    return nullptr;  // <== here
```

src/cpp/client_spawner/SentinelServerNode.h
```diff
@@ -0,0 +1,41 @@
+/**
+ * @file sentinel_server_node.h
+ * @brief Responsible for managing workers and handling their messages
+*/
+#pragma once
+
+#include "StarNode.h"
+
+#define DEBUG
```
angeloprobst: This is hard-coded but `DEBUG` is supposed to be set only while running the build process by the building tool, in our case, `Bazel`. By default, `Bazel` does not set `-DNDEBUG` (see [here](https://bazel.build/docs/user-manual#compilation-mode)), which means that `DEBUG` is always _enabled_ unless we instruct `Bazel` to not doing it.

The way we are currently running the build process does not allow us to pass options to `Bazel`, but, once we consider our services stable and ready for production, the build process will be adjusted in order to allow non-debug builds.

In short, you can, and should, remove this `#define DEBUG`.

src/cpp/client_spawner/WorkerClientNode.h
```diff
@@ -0,0 +1,27 @@
...
+namespace client_spawner {
+
+class WorkerClientNode : public StarNode {
+
+public:
+    WorkerClientNode(const string& node_id, const string& server_id, const string& das_node_server_id);
+    
+    ~WorkerClientNode();
+    
+    void execute(vector<string>& request);
```
angeloprobst:
```suggestion
    void execute(const vector<string>& request);
```

src/cpp/client_spawner/WorkerClientNode.cc
```diff
@@ -0,0 +1,65 @@
...
+    const string &das_node_server_id
+) : StarNode(node_id, server_id) {
+        this->node_id = node_id;
+        this->server_id = server_id;
+        this->das_node_server_id = das_node_server_id;
+}
+
+WorkerClientNode::~WorkerClientNode() {}
+
+void WorkerClientNode::execute(vector<string> &request) {
```
angeloprobst:
```suggestion
void WorkerClientNode::execute(const vector<string> &request) {
```

src/cpp/client_spawner/WorkerClientNode.cc
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+    }
+    
+    if (count == 0) {
+        string no_match = "No match for query";
+        cout << no_match << endl;
+        message.push_back(no_match);
+    } else {
+        cout << query_answer_str << endl;
```
angeloprobst:
```suggestion
#ifdef DEBUG
        cout << no_match << endl;
#endif
        message.push_back(no_match);
    } else {
#ifdef DEBUG
        cout << query_answer_str << endl;
#endif
```

Makefile
```diff
@@ -28,6 +28,9 @@ github-runner:
 build-image:
 	cd src && bash -x scripts/docker_image_build.sh
 
+build-spawner-image:
```
angeloprobst: See comment on `src/scripts/docker_image_spawner_build.sh` file.

```suggestion
build-spawner-image: build
```

src/scripts/docker_image_spawner_build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -x
```
angeloprobst, line 1: I think the dependencies in this script should be replaced by dependencies of a target in the `Makefile`.

I mean, as the **building image** (L9) and the **build process** (L16) already have target entries in the `Makefile`, this script should only run the `docker buildx ...` command (L28) for the spawner and have its own `Makefile` target entry with its deps set accordingly.

src/cpp/client_spawner/SentinelServerNode.h
```diff
@@ -0,0 +1,41 @@
...
+constexpr const char* WORKER_NOTIFICATION = "worker_notification";
+
+class SentinelServerNode: public StarNode {
+
+public:
+    SentinelServerNode(const string &node_id);
+
+    ~SentinelServerNode();
+
+    void storage_message(string &worker_id, string &message);
```
angeloprobst: Maybe?
```suggestion
    void storage_message(const string &worker_id, const string &message);
```

Please set them `const` if the strings are not supposed to be changed inside the function.

src/cpp/client_spawner/SentinelServerNode.h
```diff
@@ -0,0 +1,41 @@
...
+    
+    virtual shared_ptr<Message> message_factory(string &command, vector<string> &args);
+};
+
+class WorkerMessage: public Message {
+
+public:
+    string worker_id;
+    string msg;
+    WorkerMessage(string &worker_id, string &msg);
```
angeloprobst:
```suggestion
    WorkerMessage(const string &worker_id, const string &msg);
```

src/cpp/client_spawner/SentinelServerNode.cc
```diff
@@ -0,0 +1,43 @@
...
+}
+
+
+// =================================
+
+WorkerMessage::WorkerMessage(string &worker_id, string &msg) {
```
angeloprobst, line 28:
```suggestion
WorkerMessage::WorkerMessage(const string &worker_id, const string &msg) {
```

src/cpp/client_spawner/SentinelServerNode.cc
```diff
@@ -0,0 +1,43 @@
...
+    this->worker_id = worker_id;
+    this->msg = msg;
+}
```
angeloprobst:
```suggestion
WorkerMessage::WorkerMessage(const string &worker_id, const string &msg)
: worker_id(worker_id), msg(msg) {}
```

src/cpp/client_spawner/WorkerClientNode.cc
```diff
@@ -0,0 +1,65 @@
...
+
+WorkerClientNode::WorkerClientNode(
+    const string &node_id,
+    const string &server_id,
+    const string &das_node_server_id
+) : StarNode(node_id, server_id) {
+        this->node_id = node_id;
+        this->server_id = server_id;
+        this->das_node_server_id = das_node_server_id;
+}
```
angeloprobst: `StarNode` is already storing `server_id` (attr has the same name), and `DistributedAlgorithmNode` (parent of `StarNode`) is also storing `node_id` as `my_node_id` (private) but accessible via `node_id()` function.
```suggestion
WorkerClientNode::WorkerClientNode(
    const string &node_id, const string &server_id, const string &das_node_server_id
) : das_node_server_id(das_node_server_id), StarNode(node_id, server_id) {}
```

src/cpp/client_spawner/WorkerClientNode.h
```diff
@@ -0,0 +1,27 @@
...
+
+private:
+    string node_id;
```
angeloprobst: Already defined in `DistributedAlgorithNode` as `my_node_id` (private) but accessible via `node_id()` function - it can removed here.
```suggestion
```

src/cpp/client_spawner/WorkerClientNode.h
```diff
@@ -0,0 +1,27 @@
...
+    string das_node_server_id;
```
angeloprobst: Is this any different from `server_id` defined in `StarNode`?

src/cpp/client_spawner/WorkerClientNode.cc
```diff
@@ -0,0 +1,65 @@
...
+    
+    QueryAnswer *query_answer;
+    
+    unsigned int count = 0;
+    
+    RemoteIterator *response = das_node_client.pattern_matcher_query(request);
+
+    vector<string> message;
+    
+    message.push_back(this->node_id);
```
angeloprobst:
```suggestion
    vector<string> message = {this->node_id()};
```

andre-senna (changes requested):
I didn't revise C++ style as @angeloprobst  and @eddiebrissow are more reliable in this regard.

But There's a design flaw in this approach which I don't think can be fixed in this PR. This flaw is not your fault, actually it's mine. The way I was envisioning the algorithm execution is not actually well coupled with our architecture. I'm already working on this. 

Regarding this PR, my suggestion is to fix it as the other revisor(s) suggested but keep it UNMERGED. You will probably need pieces of this code later on.

src/cpp/client_spawner/WorkerClientNode.h
marcocapozzoli: Yes, it's different. The das_node_server_id is used within a DasNode object that this WorkerClientNode owns. So, for a WorkerClientNode object, server_id will be its leader_id, while das_node_server_id should be the leader_id from the DasNode object. This distinction is relevant within the execute method.

marcocapozzoli: @angeloprobst 

Yes, it's different. The das_node_server_id is used within a DasNode object that this WorkerClientNode owns. So, for a WorkerClientNode object, server_id will be its leader_id, while das_node_server_id should be the leader_id from the DasNode object. This distinction is relevant within the execute method.

But I don't know if this is the best way to handle this id. I need to use it in the execute method. But, I don't know if I really need to leave it as a private parameter. Any suggestions?

PR #256, opened 2025-02-07
[#179] Move mbed dependency to bazel registry

### LINKED ISSUE #179: Fix `mbedcrypto` building dependency
Currently, `mbedcrypto` is being set up as a local repository, built from its source code, that is located in the repo under `{repo root}/src/assets/`, inside of `3rd-party.tar.gz`. 

See `{repo root}/src/docker/Dockerfile` for more details.

However, this same `Dockerfile` installs `libmbedtls14` and `libmbedtls-dev` via `apt-get`, which makes `bazel` use the Ubuntu packages to compile and link the targets instead of using the library from the local repository.

Pedrobc89:
closes #179


src/cpp/MODULE.bazel
```diff
@@ -14,17 +14,24 @@ Additionally, this file configures:
 """
```
angeloprobst, line 15: Please remove the line 12 which mentions the fix you are doing here.

src/scripts/bazel_build.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -x
 
+set -eoux pipefail
```
angeloprobst, line 3: With the addition of this line we can safely remove all `\ &&` from the lines below, and also remove the first check in this statement: `if [ "${?}" = "0" -a "$BUILD_WHEELS" = true ]; then` (L47).

angeloprobst (approved):
LGTM, just some minor change suggestions.

src/cpp/MODULE.bazel
Pedrobc89: Addressed in [c0845bb](https://github.com/singnet/das/pull/256/commits/c0845bbd0f73b37361c83afd68ecf71a8b3f2744)

src/scripts/bazel_build.sh
Pedrobc89: Addressed in [c0845bb](https://github.com/singnet/das/pull/256/commits/c0845bbd0f73b37361c83afd68ecf71a8b3f2744)

PR #267, opened 2025-02-13
[#182] Parametrizable bazel command

### LINKED ISSUE #182: Make containerized bazel parameterizable
(BTW, the coolest ticket title ever)

Current implementation has a bunch of bash scripts to run bazel in a docker container, providing as many scripts as per needed functionality. Each script instructs bazel to run a unique and specific task, there is no way to pass arguments to bazel and control what it does.

For example, there is a script to run the unit tests, which executes all unit tests found in the repository. There is no way to select one single test file to run. Same limitations happen when building targets.

The ask here is to create a "master script" which runs bazel in a container but allows us to parameterize the bazel command.

### Examples:

```bash
# building only the specified target
./src/script/bazel.sh build //hyperon_das:hyperon_das_wheel --define=DAS_VERSION=0.9.17

# running one single test file
./src/script/bazel.sh test //tests/cpp:worker_threads_test

# updating only the Python DEV requirements.txt
./src/script/bazel.sh run //deps:requirements_dev.update

# building a specific target but also setting the number of jobs in the same command
./src/script/bazel.sh build --jobs=10 //:link_creation_server

# cleaning bazel cached and built targets
./src/script/bazel.sh clean
```
Ideally, the `bazel.sh` script should accept any number of arguments, of any kind, and just forward them to `bazel`, making no validation.

Pedrobc89:
closes #182

Add a makefile instruction that runs bazel commands and accepts extra params.

To use run:

```sh
# make bazel {extra params}
make bazel build //:link_creation_server
make bazel test //tests/python/integration:test_attention_broker_gateway
```


levisingularity (approved):

angeloprobst (dismissed):

Makefile
```diff
@@ -28,11 +28,9 @@ github-runner:
...
 	@bash -x src/scripts/docker_image_build.sh
 
-build: build-image
-	@bash -x src/scripts/build.sh
-
-tests: build-image
-	@bash -x src/scripts/unit_tests.sh
+build-all: build-image
+	@echo "building"
+	# @bash -x src/scripts/build.se
```
angeloprobst: Maybe?
```suggestion
	@bash -x src/scripts/build.sh
```

src/scripts/bazel.sh
```diff
@@ -0,0 +1,45 @@
...
+
+set -eou pipefail
+
+IMAGE_NAME="das-builder"
+CONTAINER_NAME=${IMAGE_NAME}-container
+# BAZEL_CMD="echo /opt/bazel/bazelisk $@"
+
+# local paths
+LOCAL_WORKDIR=$(pwd)
+LOCAL_BIN_DIR=$LOCAL_WORKDIR/src/bin
```
angeloprobst, line 11:
```suggestion
```

src/scripts/bazel_build.sh
```diff
@@ -2,12 +2,9 @@
 
 set -eoux pipefail
 
-JOBS=$(( $(nproc) / 2 ))
-JOBS=$(( JOBS < 1 ? 1 : JOBS ))
-
 BAZELISK_CMD=/opt/bazel/bazelisk
```
angeloprobst, line 5:
```suggestion
BAZELISK_CMD=/usr/bin/bazelisk
```

src/scripts/bazel.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+set -eou pipefail
+
+IMAGE_NAME="das-builder"
+CONTAINER_NAME=${IMAGE_NAME}-container
+# BAZEL_CMD="echo /opt/bazel/bazelisk $@"
```
angeloprobst:
```suggestion
BAZEL_CMD="/opt/bazel/bazelisk"
```

src/scripts/bazel.sh
```diff
@@ -0,0 +1,45 @@
...
+  -e BIN_DIR=$CONTAINER_BIN_DIR \
+  --volume /etc/passwd:/etc/passwd:ro \
+  --volume "$LOCAL_PIP_CACHE":"$CONTAINER_PIP_CACHE" \
+  --volume "$LOCAL_PIPTOOLS_CACHE":"$CONTAINER_PIPTOOLS_CACHE" \
+  --volume "$LOCAL_BAZEL_CACHE":"$CONTAINER_BAZEL_CACHE" \
+  --volume "$LOCAL_BAZELISK_CACHE":"$CONTAINER_BAZELISK_CACHE" \
+  --volume "$LOCAL_WORKDIR":"$CONTAINER_WORKDIR" \
+  --workdir "$CONTAINER_WORKSPACE_DIR" \
+  "${IMAGE_NAME}" \
+  bazel "$@"
```
angeloprobst:
```suggestion
  ${BAZEL_CMD} "$@"
```

src/docker/Dockerfile
```diff
@@ -30,7 +30,8 @@ RUN cd ${THIRD_PARTY} \
     && rm -f 3rd-party.tgz \
     && mkdir -p ${DAS_DIR}/src/3rd-party \
     && ln -s ${THIRD_PARTY} ${DAS_DIR}/src/3rd-party \
-    && mv bazelisk ${BAZEL_DIR}
+    && mv bazelisk ${BAZEL_DIR} \
+    && ln -s /opt/bazel/bazelisk /usr/bin/bazel
```
angeloprobst: `ln -s /opt/bazel/bazelisk /usr/bin/bazel`
I don't think this is a good thing because `bazelisk` already downloads and saves the latest `bazel` binary somewhere else. The precedence of paths can cause some circular reference.

src/.bazelrc
```diff
@@ -0,0 +1,45 @@
...
+# flag is present while invoking bazel ex: bazel build --config=debug //...
+
+#################################### COMMON ###################################
+# common --keep_going
+common --color=yes
+common --curses=yes
+# common --show_timestamps
+
+# Use half of all available CPU cores
+common --jobs=HOST_CPUS*0.6
```
angeloprobst: `0.6`? Shouldn't be `0.5`?

src/.bazelrc
```diff
@@ -0,0 +1,45 @@
...
+# test --jobs=HOST_CPUS*0.5
+
+# Other options
+test --test_output=errors
+test --build_tests_only
+test --test_verbose_timeout_warnings
+test --flaky_test_attempts=1
+test --cache_test_results=auto
+test --test_env=BAZEL_TEST_ENV=1
+test --test_tag_filters=-skip
```
angeloprobst, line 42: It seems these params are being ignored when running tests.

Makefile
```diff
@@ -52,3 +49,13 @@ setup-nunet-dms:
 reset-nunet-dms:
 	@bash -x src/scripts/reset-nunet-dms.sh
 
+bazel:
+	@bash ./src/scripts/bazel.sh $(filter-out $@, $(MAKECMDGOALS))
+
+test-all: build-image
+	$(MAKE) bazel test //...
```
angeloprobst, line 56: The suggested change below is not necessary if we figure out how to set the same behavior in `.bazelrc`:
```suggestion
	$(MAKE) bazel test --test_output=errors --build_tests_only --test_tag_filters=-skip" //...
```

src/scripts/bazel.sh
Pedrobc89: Docker entrypoint tries to find the whole string (in this case `/opt/bazel/bazelisk {params passed to make}) as a file, and by doing so, the file does not exists, what we have to do is:
```
docker run \
    ... # other params
    -- entrypoint "$BAZEL_CMD" \
    "$IMAGE_NAME" \
    $@  # Extra arguments passed to make
```

src/.bazelrc
Pedrobc89: Why do you say so, I'm 100% sure that at least the `test --test_tag_filters=-skip` is working, have no other reason to believe other flags are not.

angeloprobst: Nevermind. I misinterpreted the error output [here](https://github.com/singnet/das/actions/runs/13334865881/job/37247714262?pr=267).

Makefile
angeloprobst: Resolved. See [here](https://github.com/singnet/das/pull/267#discussion_r1958062364).

angeloprobst (approved):

PR #271, opened 2025-02-18
[#209] Create links using template

eddiebrissow:


src/link_creation_agent/compose.yaml
```diff
@@ -73,7 +73,7 @@ services:
     ports:
       - 37007:37007
     volumes:
-      - ../../../src:/opt
+      - ../../src:/opt
```
angeloprobst, line 76: Using relative paths can cause some issues.
Is there any other way to do it?

eddiebrissow: I agree, I'll rewrite the compose, rn is just for local testing

angeloprobst (approved):

Issue #232, opened 2025-01-30
Implement Inference Agent


andre-senna:
Inference Agent is a server in DAS architecture (like Link Creation Agent and Query Agent). They are supposed to be in the same DASNode network.

Inference Agent processes inference requests. There will be several different types of inferences requests but for now we'll focus only in two of them:

* PROOF_OF_IMPLICATION_OR_EQUIVALENCE
* PROOF_OF_IMPLICATION
* PROOF_OF_EQUIVALENCE

Each request types should be implemented as commands which expect potentially different parameters, although, for these three commands we're going to implement, the list of parameters is the same.

```
PROOF_OF_IMPLICATION_OR_EQUIVALENCE <handle1> <handle2> <max proof length>
PROOF_OF_IMPLICATION <handle1> <handle2> <max proof length>
PROOF_OF_EQUIVALENCE <handle1> <handle2> <max proof length>
```

Inference requests are processed mostly in the same way:

1. Send request to Link Creation Agent to create reasoning links passing a list of [QUERY_LINK_CREATION, PROCESSOR/CREATION_TEMPLATE] pairs
2. Start Distributed Inference Control algorithm passing QUERY_INFERENCE and FITNESS_FUNCTION
3. Wait for the inference to finish
4. Send results back to the caller

But different types of inference will use different queries and different processors and fitness functions in steps 1. and 2.

Both, Link Creation Agent and Inference Control Algorithm will be part of DAS architecture so we can just use shared identifiers to indicate which link processor or which fitness function are supposed to be used. I suggest using string enumerations shared in `src/cpp/commons`. 

Processors will be implemented in the Link Creation Agent while fitness functions will be implemented in the Distributed Inference Control algorithm. 

Follows the sequence diagram depicting how inference requests are supposed to unfold.

![Image](https://github.com/user-attachments/assets/d1faff41-9e02-4539-bbc8-20fcc8f65cfa)

### Link Creation request

Actually, the Inference Agent issues multiples requests to the Link Creation Agent.

### (1) Requests to create SATISFYING_SET and PATTERNS

Query: 

```
LINK_TEMPLATE Expression 3
    NODE Symbol EVALUATION
    LINK_TEMPLATE Expression 2
        NODE Symbol PREDICATE
        VARIABLE P
    LINK_TEMPLATE Expression 2
        NODE Symbol CONCEPT
        VARIABLE C
```

Creation template:

```
LIST 2
    LINK_CREATE Expression 2 1
[...]

PR #281, opened 2025-02-27
[#232] Inference Agent implementation 2/3

eddiebrissow:
links to #232 

Inference flow control;
Missing inference request implementation (Any query string and template string hardcoded will be replaced in the next PR);


src/commons/Utils.cc
```diff
@@ -102,3 +103,23 @@ string StopWatch::str_time() {
         return to_string(millis) + " millis";
     }
 }
+
+map<string, string> Utils::parse_config(string const &config_path) {
+    map<string, string> config;
+    ifstream file(config_path);
+    string line;
+    while (getline(file, line)) {
+        istringstream is_line(line);
```
angeloprobst: Pls rename this var to `iss` (or `line_iss`, or similar) because `is_line` resembles a boolean flag which means `is it a line?`.

src/commons/Utils.cc
```diff
@@ -102,3 +103,23 @@ string StopWatch::str_time() {
         return to_string(millis) + " millis";
     }
 }
+
+map<string, string> Utils::parse_config(string const &config_path) {
+    map<string, string> config;
+    ifstream file(config_path);
```
angeloprobst, line 109: This file seems to be a list of `key = value` entries.
Is there any example file in the repo?

angeloprobst (approved):
Minor comment to be addressed.
Approved.

src/commons/Utils.cc
eddiebrissow: done

src/commons/Utils.cc
eddiebrissow: There is an example in the README, no config file

PR #285, opened 2025-03-05
[#209] Link Creation Agent support to list of templates

eddiebrissow:
Added support to list in Link Creation Agent
Commented no supported commands

### DIFF
--- a/src/link_creation_agent/das_agent_node.cc
+++ b/src/link_creation_agent/das_agent_node.cc
@@ -11,6 +11,7 @@ DasAgentNode::~DasAgentNode() { DistributedAlgorithmNode::graceful_shutdown(); }
 void DasAgentNode::create_link(vector<string>& request) {
     cout << "Creating link" << endl;
     for (string token : request) {
-        cout << token << endl;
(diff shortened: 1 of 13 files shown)


src/link_creation_agent/das_agent_node.cc
```diff
@@ -11,6 +11,7 @@ DasAgentNode::~DasAgentNode() { DistributedAlgorithmNode::graceful_shutdown(); }
 void DasAgentNode::create_link(vector<string>& request) {
     cout << "Creating link" << endl;
     for (string token : request) {
-        cout << token << endl;
+        cout << token << " ";
```
angeloprobst, line 15: Only `cout`? Maybe you missed a `TODO:` here as a reminder to implement the actual feature?

angeloprobst (approved):
Minor comment.

src/link_creation_agent/das_agent_node.cc
eddiebrissow: I'll add, DAS Agent need to be implemented

PR #295, opened 2025-03-10
[198] Remove glob's on BUILD files

### LINKED ISSUE #198: Break targets that define `srcs` as `glob` or `filegroup` into single-file targets
There are some cases like this:
```python
filegroup(
    name = "py_files",
    srcs = glob(["*.py"]),
)

py_library(
    name = "hyperon_das",
    srcs = [":py_files"],
    deps = [...]
)
```
Where a single target globs all source files.

This does not help `bazel` to make the best decision about doing or not a re-build or a re-test of cached targets. Usually, this leads `bazel` to do more re-builds and re-tests than necessary.

Ideally, each source code file should have its own target and be listed in the `deps` parameter of who depends on it.

NOTE: same can be found in the CPP code, not only in Python targets.

Pedrobc89:
resolves #198

Removed almost all glob and filegroup from BUILD files, listing it's dependencies instead.
Couldn't refactor distributed_algorithm_node, due to its circular imports nature.
Couldn't remove the grouping of python files, whenever done that, we could not use the import of python packages.


angeloprobst (approved):
### Looks good.

Ran some local checks:
1. Wiped bazel cache out (`make bazel clean`).
2. Deleted bins/wheels (`rm -f ./src/bin/*`).
3. Build all (`make build-all`).
4. Executed tests (`make test-all`).
5. Spawned Attention Broker and Query Agent and executed some queries.
6. Created a fresh Python 3.10 venv, installed `hyperon-das-atomdb-cpp` wheel, imported it in Python console, instantiated nodes and save them into the DB, fetched them back.

### Everything worked as expected.

PR #296, opened 2025-03-10
[#221] non-distributed optimization algorithm

marcocapozzoli:


src/evolution/asset/hyperon_das_node-0.0.1-cp310-abi3-manylinux_2_28_x86_64.whl
andre-senna, line 1: Is this binary file really supposed to be committed?

src/evolution/fitness_functions.py
```diff
@@ -0,0 +1,26 @@
+import enum
+import math
+
+
+class FitnessFunctionsTAG(str, enum.Enum):
+    MULTIPLY_STRENGTHS = "multiply_strengths"
+
+
+def handle_fitness_function(tag: str) -> callable:
```
andre-senna: please rename to compute_fitness_function() if this function is supposed to be called directly or fitness_function_handle() if the idea is to use it to set a variable and call the function indirectly.

This or just rename it to fitness_function().

src/evolution/fitness_functions.py
```diff
@@ -0,0 +1,26 @@
...
+        raise ValueError("Unknown fitness function TAG")
+
+
+def multiply_strengths(atom_db, query) -> float:
+    strengths_value = []
+    for handle in query.handles:
+        atom = atom_db.get_atom(handle)
+        if (custom_attributes := atom.custom_attributes):
+            if (strength := custom_attributes.get('strength')):
+                strengths_value.append(strength)
```
andre-senna: Why storing strength values in a list? Why not just compute the product?

src/evolution/optimizer.py
```diff
@@ -0,0 +1,363 @@
...
+# However, it is necessary to create bindings for the components already implemented in C++
+# and integrate them into this implementation.
+# Currently, I am using the implementation from this PR:
+from evolution.das_node.query_answer import QueryAnswer
+from evolution.das_node.das_node import DASNode
+
+MAX_CORRELATIONS_WITHOUT_STIMULATE = 1000
+
+
+class SharedQueue:
```
andre-senna: I didn't understand why you are wrapping a Python queue inside this new class. What does this class add?

src/evolution/selection_methods.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Returns:
+        callable: Function that performs the selection of individuals.
+
+    Raises:
+        ValueError: If the provided selection method is unknown.
+    """
+    if method == SelectionMethodType.BEST_FITNESS:
+        return best_fitness
+    elif method == SelectionMethodType.FITNESS_PROPORTIONATE:
```
andre-senna: Do you mean "FITNESS_PROPORTIONAL"?  Anyway, even "proportional" is an awful name :-P

Please use ROULETTE instead, which is a more usual name for this function. However, I have two comments which don't need to be addressed in this PR but need to be fixed (please create separate cards for them).

Firstly, this roulette selection method allows the same individual to be selected more than once. Although this is OK in some implementations of GA/GP, in the case of our optimization algorithm tit doesn't seem to make sense IMO.

Secondly, we should implement a tournament selection as well. It's supposed to give equivalent results (when compared to roulette selection) but it's a lot cheaper.

andre-senna (changes requested):

src/evolution/asset/hyperon_das_node-0.0.1-cp310-abi3-manylinux_2_28_x86_64.whl
angeloprobst: This python module is already being published on PyPI: https://pypi.org/project/hyperon-das-node/.
So, a `pip install hyperon-das-node==0.0.1` might work.

src/evolution/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ENV PYTHONPATH=/app
+
+RUN apt-get update && apt-get install -y git
+
+RUN git clone https://github.com/singnet/das-query-engine.git
+RUN cd das-query-engine && git checkout ${DAS_QUERY_ENGINE_BRANCH} &&  mv hyperon_das/ ../hyperon_das
+RUN pip3 install --no-cache --upgrade pip hyperon_das_atomdb requests grpcio google protobuf pyinstaller
+
+COPY asset/hyperon_das_node-0.0.1-cp310-abi3-manylinux_2_28_x86_64.whl /app/
+RUN pip3 install /app/hyperon_das_node-0.0.1-cp310-abi3-manylinux_2_28_x86_64.whl
```
angeloprobst: See https://github.com/singnet/das/pull/296/files#r1991458637.

src/evolution/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+pyinstaller --onefile ./evolution/main.py
+mkdir -p bin
+mv ./dist/main ./dist/evolution
+mv ./dist/evolution ./bin/evolution
+rm -rf build/ __pycache__
```
angeloprobst, line 7: Is this creating a `wheel` file?

Please check how the other modules are created using a `bazel` target: https://github.com/singnet/das/blob/ed760b3616474a448c69db94694ada95d6e45f76/src/hyperon_das/BUILD#L35-L58

src/evolution/fitness_functions.py
```diff
@@ -0,0 +1,26 @@
...
+            # if (truth_value := custom_attributes.get('truth_value')):
+            #     strength, confidence = truth_value
+            #     strengths_value.append(strength)
+    return math.prod(strengths_value)
```
angeloprobst: I think this is cleaner/simpler and also addresses @andre-senna's comment.
Unless I'm not seeing other problems/issues this change could cause.

```python
import math
from typing import ClassVar

def multiply_strengths(atom_db, query) -> float:
    strengths_value = []
    for handle in query.handles:
        atom = atom_db.get_atom(handle)
        if (custom_attributes := atom.custom_attributes):
            if (strength := custom_attributes.get('strength')):
                strengths_value.append(strength)
            # if (truth_value := custom_attributes.get('truth_value')):
            #     strength, confidence = truth_value
            #     strengths_value.append(strength)
    return math.prod(strengths_value)

class FitnessFunctions:
    _fitness_functions: ClassVar[dict[str, callable]] = {
    	"multiply_strengths": multiply_strengths
    }

	@classmethod
    def get(cls, fitness_function_name: str) -> callable:
        if fitness_function_name not in cls._fitness_functions:
            raise ValueError("Unknown fitness function")
        return cls._fitness_functions[fitness_function_name]
```

In the places where it is used (just an example):
```python
from fitness_functions import FitnessFunctions
...
class QueryOptimizerIterator:
    ...
    def _evaluate(self, query_answer: QueryAnswer) -> tuple[QueryAnswer, float]:
        """ ... """
        fitness_function = FitnessFunctions.get(self.fitness_function)
        ...
```

src/evolution/optimizer.py
angeloprobst: It seems you created this wrapper just to have a `ShareQueue` with the same interface/signatures as the C++ implementation. Any reason for that?

Another question is why there is another quite similar implementation in this PR: https://github.com/singnet/das/blob/masc/221-non-distributed-optimization-algorithm/src/evolution/das_node/shared_queue.py

src/evolution/selection_methods.py
angeloprobst: I think `cycle` from `itertools` can be useful here: https://docs.python.org/3/library/itertools.html#itertools.cycle

angeloprobst: Also, this file could follow the same pattern I've suggested for `FitnessFunctions`.

src/evolution/build.sh
marcocapozzoli: This does not create a wheel file.

src/evolution/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:22.04
+
+ARG DAS_QUERY_ENGINE_BRANCH="masc/change-service-in-grpc"
+
+RUN apt-get update && apt-get install -y \
+    python3.10 \
+    python3.10-distutils \
+    python3.10-venv \
+    python3-pip \
+    binutils \
```
angeloprobst, line 10:
```suggestion
    binutils \
    git \
```

src/evolution/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1
+
+WORKDIR /app
+
+ENV PYTHONPATH=/app
+
+RUN apt-get update && apt-get install -y git
```
angeloprobst:
```suggestion
```

src/evolution/build.sh
angeloprobst: Oh, got it. It's a way of creating a kind of a binary which embeds a python application along with its dependencies, allowing this:
https://github.com/singnet/das/blob/2bcf58823185a06f8bd009c0e5c852919b125472/Makefile#L52-L53

I suppose this is temporary and will be replaced by a `wheel` file once it is considered ready/done, right?

src/evolution/asset/hyperon_das_node-0.0.1-cp310-abi3-manylinux_2_28_x86_64.whl
marcocapozzoli: I uploaded this binary just to make it easier to create the environment, but I had forgotten that we had this library on Pypi. So, I did what Angelo suggested.

src/evolution/fitness_functions.py
marcocapozzoli: Ok. But I applied what Angelo suggested in a comment below. So this function no longer exists.

src/evolution/fitness_functions.py
marcocapozzoli: I changed the approach and removed the list

src/evolution/optimizer.py
marcocapozzoli: This class doesn't add anything, it was just a way to organize the code. I removed the class

src/evolution/Dockerfile
marcocapozzoli: Ok

src/evolution/selection_methods.py
marcocapozzoli: I changed the function to not allow the same individual to be selected more than once

andre-senna: I don't think `cycle` does what we need here. The "Roulette" algorithm is not like a round robin selection. The idea is to select a individual with a probability which is proportional to its fitness value.

andre-senna (approved):

src/evolution/build.sh
marcocapozzoli: > It's a way of creating a kind of a binary which embeds a python application along with its dependencies
> I suppose this is temporary and will be replaced by a wheel file once it is considered ready/done, right?

It's a way to test the algorithm in a simpler way. This binary probably won't exist, but I don't know if we would have a wheel file up front. In short, yes, this is temporary.

PR #299, opened 2025-03-12
[291] Nanobind for QueryAnswer and DASNode

### LINKED ISSUE #291: Expose QueryEngine to Python via nanobind
At first we only need to expose DasNode and QueryAnswer.

Pedrobc89:
closes #291


src/hyperon_das_query_engine/hyperon_das_query_engine_ext.cc
```diff
@@ -0,0 +1,85 @@
...
+    .def("to_string", &QueryAnswer::to_string);
+
+  //HandlesAnswer.h
+  nb::class_<HandlesAnswer, QueryAnswer>(m, "HandlesAnswer")
+    .def(nb::init<>())
+    .def(nb::init<double>(), "importance"_a)
+    .def(nb::init<const char*, double>(), "handle"_a, "importance"_a)
+    //TODO: nanobind is failing to bind the handles field
+    // error: invalid conversion from 'const char* const*' to 'char'
+    // .def_ro("handles", &HandlesAnswer::handles)
```
angeloprobst: @pedro-costa-techlabs `const char*` (or any other type of pointer) in nanobind is always challenging and should be avoided.

This particular case is even more complicated because `handles` is an array of pointers, so the binding implementation must translate it to a format that nanobind knows how to deal with, as follows:

```cpp
#include <nanobind/stl/vector.h>
...
    .def_prop_ro(
        "handles",
        [](const HandlesAnswer& self) -> const vector<string> {
            vector<string> handles;
            for(size_t i = 0; i < self.handles_size; i++) {
                handles.emplace_back(self.handles[i]);
            }
            return handles;
        })
```

src/hyperon_das_query_engine/hyperon_das_query_engine_ext.cc
```diff
@@ -0,0 +1,85 @@
...
+  nb::class_<QueryAnswer>(m, "QueryAnswer")
+    .def("tokenize", &QueryAnswer::tokenize)
+    .def("untokenize", &QueryAnswer::untokenize, "tokens"_a)
+    .def("to_string", &QueryAnswer::to_string);
+
+  //HandlesAnswer.h
+  nb::class_<HandlesAnswer, QueryAnswer>(m, "HandlesAnswer")
+    .def(nb::init<>())
+    .def(nb::init<double>(), "importance"_a)
+    .def(nb::init<const char*, double>(), "handle"_a, "importance"_a)
```
angeloprobst: I think this could also require a special treatment for `const char* handle`:

```cpp
    .def(
        "__init__",
        [](HandlesAnswer& self, const string& handle, double importance) {
            new (&self) HandlesAnswer(handle.c_str(), importance);
        },
        "handle"_a,
        "importance"_a)
```

src/hyperon_das_query_engine/hyperon_das_query_engine_ext.cc
```diff
@@ -0,0 +1,85 @@
...
+    .def_ro("handles_size", &HandlesAnswer::handles_size)
+    .def_ro("importance", &HandlesAnswer::importance)
+    .def("add_handle", &HandlesAnswer::add_handle, "handle"_a)
+    .def("merge", &HandlesAnswer::merge, "other"_a, "merge_handles"_a)
+    .def_static("copy", &HandlesAnswer::copy, "base"_a)
+  ;
+
+  // RemoteIterator.h
+  nb::class_<RemoteIterator<HandlesAnswer>>(m, "RemoteIterator")
+   .def(nb::init<string>(), "local_id"_a)
```
angeloprobst:
```suggestion
   .def(nb::init<const string&>(), "local_id"_a)
```

Please revisit the other functions/constructors to ensure that the binding is expecting/passing in the right types, including aspect modifiers like `const`, `&`, etc.

src/hyperon_das_query_engine/hyperon_das_query_engine_ext.cc
angeloprobst: Done: [`18c72c0` (#299)](https://github.com/singnet/das/pull/299/commits/18c72c0e64e74cf01f7ae71d068039e1b81d3487)

src/hyperon_das_query_engine/hyperon_das_query_engine_ext.cc
angeloprobst: Done: [`18c72c0` (#299)](https://github.com/singnet/das/pull/299/commits/18c72c0e64e74cf01f7ae71d068039e1b81d3487)

src/hyperon_das_query_engine/hyperon_das_query_engine_ext.cc
angeloprobst: Done: [`18c72c0` (#299)](https://github.com/singnet/das/pull/299/commits/18c72c0e64e74cf01f7ae71d068039e1b81d3487)

Issue #206, opened 2025-01-24
Fix processing of LinkTemplates with large number of answers


andre-senna:
There's already a "pagination" scheme in place but it's not working properly and need to be fixed or re-designed.

arturgontijo:
Resolved by #325

PR #301, opened 2025-03-13
[#206] Use ZRANGE to fetch paginated data from redis.

arturgontijo:
Solves #206 by introducing the usage of Redis Sorted Set.

The chunk_size (page_size) is set to `10,000` but we can change it by setting `DAS_REDIS_CHUNK_SIZE`.

Deps:
  - https://github.com/singnet/das-metta-parser/pull/63


andre-senna (approved):

src/query_engine/AtomDB.cc
```diff
@@ -140,16 +140,56 @@ void AtomDB::mongodb_setup() {
...
     }
-    // NOTE: Intentionally, we aren't destroying 'reply' objects.'reply' objects are destroyed in ~RedisSet().
-    return shared_ptr<atomdb_api_types::RedisSet>(new atomdb_api_types::RedisSet(reply));
+
+    this->redis_current_chunk = shared_ptr<atomdb_api_types::HandleList>(new atomdb_api_types::RedisSet(reply));
+
+    this->redis_has_more = (reply->elements == this->redis_chunk_size);
+}
+
+vector<std::shared_ptr<atomdb_api_types::HandleList>> AtomDB::query_for_pattern(std::shared_ptr<char> pattern_handle) {
```
angeloprobst: I think a single function doing all the job that is made here by these 2 methods is cleaner and simpler.

Also, we should avoid creating new object instance attributes where local variables could do the job.

angeloprobst: Single function implementation:
```cpp
vector<shared_ptr<atomdb_api_types::HandleList>> AtomDB::query_for_pattern(std::shared_ptr<char> pattern_handle) {
    unsigned int redis_cursor = 0;
    bool redis_has_more = true;
    unsigned int redis_chunk_size = REDIS_CHUNK_SIZE;
    string command;
    redisReply *reply;
    shared_ptr<atomdb_api_types::HandleList> redis_chunk;

    string chunk_size_str = Utils::get_environment("DAS_REDIS_CHUNK_SIZE");
    if (chunk_size_str != "") {
        redis_chunk_size = stoi(chunk_size_str);
    }

    vector<std::shared_ptr<atomdb_api_types::HandleList>> results;
    
    while (redis_has_more) {
        command = (
            "ZRANGE "
            + REDIS_PATTERNS_PREFIX
            + ":"
            + pattern_handle.get()
            + " "
            + to_string(redis_cursor)
            + " "
            + to_string(redis_cursor + redis_chunk_size - 1)
        );
        
        reply = (redisReply *) redisCommand(this->redis_single, command.c_str());

        if (reply == NULL) {
            Utils::error("Redis error");
            break;
        }

        if (reply->type != REDIS_REPLY_SET && reply->type != REDIS_REPLY_ARRAY) {
            Utils::error("Invalid Redis response: " + std::to_string(reply->type));
            freeReplyObject(reply);
            break;
        }

        redis_chunk = shared_ptr<atomdb_api_types::HandleList>(new atomdb_api_types::RedisSet(reply));
        
        if (redis_chunk) {
            results.push_back(redis_chunk);
        }

        redis_cursor += redis_chunk_size;
        redis_has_more = (reply->elements == redis_chunk_size);
    }

    return results;
}
```

angeloprobst: `REDIS_CHUNK_SIZE` must be defined at the top of `AtomDB.h`:
```cpp
...
using namespace std;

#define REDIS_CHUNK_SIZE ((unsigned int) 10000)

namespace query_engine {
...
```

angeloprobst: And then remove these attributes from `AtomDB.h`:
```cpp
    shared_ptr<char> redis_pattern_handle;
    unsigned int redis_chunk_size = 10000;
    unsigned int redis_cursor = 0;
    bool redis_has_more = true;
    std::shared_ptr<atomdb_api_types::HandleList> redis_current_chunk;
```

src/query_engine/query_element/LinkTemplate.h
```diff
@@ -270,18 +270,25 @@ class LinkTemplate : public Source {
         cout << "fetch_links() Pattern handle: " << this->handle << endl;
 #endif
         shared_ptr<AtomDB> db = AtomDBSingleton::get_instance();
-        this->fetch_result = db->query_for_pattern(this->handle);
-        unsigned int answer_count = this->fetch_result->size();
+        this->fetch_results = db->query_for_pattern(this->handle);
```
angeloprobst: I'm wondering if the return of `AtomDB::query_for_patten` should be changed to be a list of strings and avoid the double-for-loop statements in this function.

Loops to read the handles and put them together in a single list would still be required, but would happen inside of `AtomDB::query_for_pattern`, not here.

@andre-senna WDYT?

src/query_engine/query_element/LinkTemplate.h
```diff
@@ -270,18 +270,25 @@ class LinkTemplate : public Source {
...
         vector<HandlesAnswer*> fetched_answers;
         if (answer_count > 0) {
             dasproto::HandleList handle_list;
             handle_list.set_context(this->context);
-            for (unsigned int i = 0; i < answer_count; i++) {
-                handle_list.add_list(this->fetch_result->get_handle(i));
+            for (unsigned int i = 0; i < vector_len; i++) {
+                for (unsigned int j = 0; j < this->fetch_results[i]->size(); j++) {
+                    handle_list.add_list(this->fetch_results[i]->get_handle(j));
+                }
```
angeloprobst: Yikes!

src/query_engine/query_element/LinkTemplate.h
```diff
@@ -293,25 +300,28 @@ class LinkTemplate : public Source {
...
-                        Terminal* terminal = (Terminal*) this->target_template[j];
-                        if (terminal->is_variable) {
-                            if (!query_answer->assignment.assign(
-                                    terminal->name.c_str(), this->atom_document[i]->get("targets", j))) {
-                                Utils::error(
-                                    "Error assigning variable: " + terminal->name +
-                                    " a value: " + string(this->atom_document[i]->get("targets", j)));
+            unsigned int idx = 0;
+            for (unsigned int i = 0; i < vector_len; i++) {
+                for (unsigned int j = 0; j < this->fetch_results[i]->size(); j++) {
```
angeloprobst: Yikes!

src/query_engine/query_element/LinkTemplate.h
```diff
@@ -270,18 +270,25 @@ class LinkTemplate : public Source {
         cout << "fetch_links() Pattern handle: " << this->handle << endl;
 #endif
         shared_ptr<AtomDB> db = AtomDBSingleton::get_instance();
-        this->fetch_result = db->query_for_pattern(this->handle);
-        unsigned int answer_count = this->fetch_result->size();
+        this->fetch_results = db->query_for_pattern(this->handle);
+        unsigned int vector_len = this->fetch_results.size();
+        unsigned int answer_count = 0;
+        for (unsigned int i = 0; i < vector_len; i++) {
+            answer_count += this->fetch_results[i]->size();
```
angeloprobst: Yikes!

src/query_engine/query_element/LinkTemplate.h
andre-senna: @angeloprobst  @arturgontijo 
Yes, I think you're right. It's better to make it inside query_for_pattern() which is where all the hacks are happening anyway.

src/query_engine/AtomDB.cc
arturgontijo: Cool, done at [dff2dd5](https://github.com/singnet/das/pull/301/commits/dff2dd59454300e3e99f03c502d804f5270ab66b)

src/query_engine/query_element/LinkTemplate.h
arturgontijo: Done at [dff2dd5](https://github.com/singnet/das/pull/301/commits/dff2dd59454300e3e99f03c502d804f5270ab66b)

angeloprobst (approved):
Looks good.

src/query_engine/AtomDB.cc
```diff
@@ -140,16 +140,56 @@ void AtomDB::mongodb_setup() {
...
-    if (reply == NULL) {
-        Utils::error("Redis error");
+vector<string> AtomDB::query_for_pattern(std::shared_ptr<char> pattern_handle) {
+    unsigned int redis_cursor = 0;
+    bool redis_has_more = true;
+    unsigned int redis_chunk_size = REDIS_CHUNK_SIZE;
+    string command;
+    redisReply *reply;
+
+    string chunk_size_str = Utils::get_environment("DAS_REDIS_CHUNK_SIZE");
```
andre-senna: This shouldn't be an environment variable. This parameter has the potential to be changed during runtime in order to change the behavior of the AtomDB. Please make it a static class uint variable initialized with a hard-wired value.

src/query_engine/AtomDB.cc
```diff
@@ -140,16 +140,56 @@ void AtomDB::mongodb_setup() {
...
+            freeReplyObject(reply);
+            break;
+        }
+
+        for (unsigned int i = 0; i < reply->elements; i++) {
+            results.push_back(reply->element[i]->str);
+        }
+
+        redis_cursor += redis_chunk_size;
+        redis_has_more = (reply->elements == redis_chunk_size);
```
andre-senna: This condition seems wrong. If we have exactly redis_chunk_size elements stored in a given key redis_has_more will be set true and break in line 172.

Perhaps the proper way to fix is changing the check in line 171 to accept "reply == NULL" if it isn't the first time that ZRANGE is executed.

andre-senna (approved):

src/query_engine/AtomDB.cc
angeloprobst: I understood that:
1) `reply` will be `NULL` only when `redisCommand` fails.
2) if there is no more results to retrieve, `redisCommand` will return a valid but empty `redisReply` - not `NULL`.

src/query_engine/AtomDB.cc
arturgontijo: Done 403f32bc0d5647fdd5654ee96e89a148c25f9a2b

src/query_engine/AtomDB.cc
```diff
@@ -140,16 +141,50 @@ void AtomDB::mongodb_setup() {
...
+        reply = (redisReply *) redisCommand(this->redis_single, command.c_str());
+
+        if (reply == NULL) {
+            Utils::error("Redis error");
+            break;
+        }
+
+        if (reply->type != REDIS_REPLY_SET && reply->type != REDIS_REPLY_ARRAY) {
+            Utils::error("Invalid Redis response: " + std::to_string(reply->type));
+            freeReplyObject(reply);
```
angeloprobst, line 173: `reply` must be freed before calling `Utils::error` b/c it throws an exception:
```suggestion
			auto reply_type = std::to_string(reply->type);
            freeReplyObject(reply);
            Utils::error("Invalid Redis response: " + reply_type);
```

PR #302, opened 2025-03-13
[#232] Inference Agent tests

eddiebrissow:
Inference Agent tests;
Minor fixes;
links to #232

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,9 +61,6 @@ bazel:
 test-all: build-image
 	@$(MAKE) bazel test //...
 
-single-test: build-image
-	@$(MAKE) bazel test $(filter-out $@, $(MAKECMDGOALS))
-
 lint-all:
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -61,6 +61,9 @@ bazel:
 test-all: build-image
 	@$(MAKE) bazel test //...
 
+single-test: build-image
+	@$(MAKE) bazel test $(filter-out $@, $(MAKECMDGOALS))
```
angeloprobst: Why is this shortcut necessary since `make bazel <command> [<params>] <target>` is already available?

For example, `make bazel test //tests/cpp:message_broker_test` works.

angeloprobst (approved):

Makefile
eddiebrissow: Agree

PR #304, opened 2025-03-18
[#232] Added context and ID support

eddiebrissow:
Added context support to Inference Agent;
Added ID to Link Create Agent;


src/inference_agent/inference_agent.h
```diff
@@ -81,6 +81,7 @@ class InferenceAgent {
      */
     const string get_next_iterator_id();
     // Private variables
+    const string get_next_inference_request_id();
```
andre-senna, line 83: I suppose you want to add this before the comment "Private variables"

src/inference_agent/inference_request_validator.h
```diff
@@ -25,7 +25,7 @@ class InferenceRequestValidator : public RequestValidator {
    protected:
     const string inference_request_validator_regex =
```
andre-senna, line 29: It's OK to keep this string initialization here but you should compile this regex ion the agent constructor and store the compiled regex as well.

andre-senna (approved):

src/inference_agent/inference_agent.h
eddiebrissow: Done

src/inference_agent/inference_request_validator.h
eddiebrissow: Added regex compilation in the constructor

PR #307, opened 2025-03-18
[#232] Inference -> Evolution communication

eddiebrissow:
Added DAS Agent (receives a link request and create using AtomDB);
Inference communication to Evolution Aka (Distributed Inference Algorithm)
Fixed bugs;
Improvements;


Makefile
```diff
@@ -52,6 +52,9 @@ run-inference-agent-client:
 run-evolution:
 	@bash ./src/scripts/bazel.sh run //evolution:main -- $(OPTIONS)
 
+run-das-agent:
```
andre-senna, line 55: das-agent is misleading. Query agent, evolution agent, link-creation agent, inference agent... All of them ARE DAS agents.

My suggestion is das-server.

andre-senna (approved):
Approved but please make the suggested change before merging

src/das_agent/das_agent_node.py
```diff
@@ -0,0 +1,81 @@
...
+class DASLinkCreateMessage(Message):
+    def __init__(self, command: str, args: List[str]):
+        super().__init__()
+        self.message = args
+
+    def act(self, node, *args, **kwargs):
+        node.add_request(self.message)
+
+
+class StarNode(DistributedAlgorithmNode):
```
marcocapozzoli: There is no need to create a StarNode. In the hyperon_das_node package you can import the StarNode. Pedro created Bind and I am using it in evolution, it works normally. Small details are different, like is_server not existing, the correct one is is_leader()

`from hyperon_das_node import StarNode`

marcocapozzoli (changes requested):

src/das_agent/das_agent_node.py
eddiebrissow: Done

Makefile
eddiebrissow: I'll change the name in the next PR due to my current work is on stash and relies on the name.

PR #310, opened 2025-03-20
[#209] New Link Creation Agent processors

### LINKED ISSUE #209: Implement link creation algorithm

eddiebrissow:
Key points/files:

-  `equivalence_processor.h`
    -  the `get_pattern_query` method is a union set between two concepts
    - `process` method contains all logic to create the EQUIVALENCE links
- `implication_processor.cc`
   - `get_pattern_query` query to count all occurrencies of a predicate
   - `get_satisfying_set_query` an intersection between two predicates
   - `process` method contains all logic to create the IMPLICATION links
- `link_processor.h`
  - `_count_query` and `count_query`:  send count queries to QA and wait to finish, there is a post delay to ensure stability
- `link_creation_service.cc`
  - All the logic to create links given an initial query 

closes #209


andre-senna (changes requested):
See comments in the dev channel

andre-senna (changes requested):
There are two things in this new design that I think new to be changed.

(1) Processor are designed to deal with 1 single QueryAnswer. This is bad because we'd need to instantiate one processor for each QueryAnswer and this is a totally unnecessary overhead. And a pointless one, in my opinion, since it doesn't make the design simpler. Processors don't keep any data (or they shouldn't) so the best design here is to instantiate them only once in agent's constructor and use them to process QueryAnswers of multiple requests.

(2) Creation and returning of the created links are split into three methods making the whole process asynchronous. This introduce unnecessary overhead and complexity. I believe the need to tho this is because of the point (1) above. So I think that solving (1) will make it easier to solve (2).

NOTE: to address (1) and (2) we'd need to change the method signature to something like:

```
    virtual virtual std::vector<std::vector<std::string>> process(QueryAnswer* query_answer);
```

Which kind of merge the functionality of current constructor, process() and get_links().

src/link_creation_agent/service.cc
```diff
@@ -23,16 +24,17 @@ void LinkCreationService::process_request(shared_ptr<RemoteIterator<HandlesAnswe
...
+                if (LinkCreationProcessor::get_processor_type(link_template.front()) == 
+                ProcessorType::PROOF_OF_IMPLICATION) {
+                    link_tokens = implication_processor.process(query_answer, link_template);
+                }else if (LinkCreationProcessor::get_processor_type(link_template.front()) ==
+                ProcessorType::PROOF_OF_EQUIVALENCE) {
+                    link_tokens = equivalence_processor.process(query_answer, link_template);
+                }else {
+                    link_tokens = link_template_processor.process(query_answer, link_template);
                 }
+                this->create_link(link_tokens, *das_client);
```
angeloprobst:
```cpp
                vector<vector<string>> link_tokens;
                auto processor_type = \
                    LinkCreationProcessor::get_processor_type(link_template.front());
                if (processor_type == ProcessorType::PROOF_OF_IMPLICATION) {
                    link_tokens = \
                        implication_processor.process(query_answer, link_template);
                } else if (processor_type == ProcessorType::PROOF_OF_EQUIVALENCE) {
                    link_tokens = \
                        equivalence_processor.process(query_answer, link_template);
                } else {
                    link_tokens = \
                        link_template_processor.process(query_answer, link_template);
                }
                this->create_link(link_tokens, *das_client);
```

src/link_creation_agent/template_processor.cc
```diff
@@ -0,0 +1,59 @@
...
+#include <iostream>
+
+using namespace std;
+using namespace query_engine;
+using namespace link_creation_agent;
+
+std::vector<std::vector<std::string>> LinkTemplateProcessor::process(
+    QueryAnswer* query_answer, std::optional<std::vector<std::string>> config) {
+        std::vector<shared_ptr<Link>> links;
+        if(config != std::nullopt){
```
angeloprobst: It's an `optional` but cannot be `nullopt`?

src/link_creation_agent/template_processor.cc
```diff
@@ -0,0 +1,59 @@
...
+                    links.push_back(link);
+                }
+            } else {
+                LinkCreateTemplate link_template(config.value());
+                shared_ptr<Link> link = process_template_request(query_answer, link_template);
+                links.push_back(link);
+            }
+        } else{
+            throw runtime_error("Invalid link template");
+        }
```
angeloprobst:
```cpp
        if (config == std::nullopt) {
            throw runtime_error("Invalid link template");
        }
        if (config.value().front() == "LIST") {
            LinkCreateTemplateList link_create_template_list(config.value());
            for (auto link_template : link_create_template_list.get_templates()) {
                auto link = \
                    process_template_request(query_answer, link_template);
                links.push_back(link);
            }
        } else {
            LinkCreateTemplate link_template(config.value());
            auto link = \
                process_template_request(query_answer, link_template);
            links.push_back(link);
        }
```

angeloprobst:
Please, check the functions params for _const-ness_ and _referencing_.

Also, if possible, add `using namespace std;` to the top of the file and remove `std::` from the middle of the code to make it cleaner.

src/link_creation_agent/equivalence_processor.cc
```diff
@@ -0,0 +1,133 @@
...
+                    c2
+        };
+    // clang-format on
+    vector<string> pattern_query;
+    for (const auto& q_template : templates) {
+        for (const auto& token : q_template) {
+            pattern_query.push_back(token);
+        }
+    }
+    return pattern_query;
```
andre-senna: Considering that `templates` is hardwired, you should pre-compute and store `pattern_query` instead. I see no reason to iterate on `template` every call.

src/link_creation_agent/equivalence_processor.cc
```diff
@@ -0,0 +1,133 @@
...
+                    c1,
+            {"LINK_TEMPLATE", "Expression", "3",
+                "NODE", "Symbol", "EVALUATION",
+                "LINK_TEMPLATE", "Expression", "2",
+                    "NODE", "Symbol", "PREDICATE",
+                    "VARIABLE", "P",
+                "LINK", "Expression", "2",
+                    "NODE", "Symbol", "CONCEPT"},
+                    c2
+        };
```
andre-senna: The brackets don't seem to match properly. Please review. It seems like the opening in line 17, for instance, is closing in line 24.

src/commons/Utils.cc
```diff
@@ -136,12 +136,24 @@ std::string Utils::join(const std::vector<std::string>& tokens, char delimiter)
...
         result += charset[rand() % max_index];
     }
     return result;
 }
+
+bool Utils::is_number(const string& s) {
+    return !s.empty() &&
+           find_if(s.begin(), s.end(), [](unsigned char c) { return !isdigit(c); }) == s.end();
+}
+// TODO move this to a utils file
```
angeloprobst: Done?

src/das_agent/main.py
```diff
@@ -16,21 +18,39 @@ def get_request(node: DASAgentNode):
 def parse_targets(targets, atom_db):
     new_targets = []
     for target in targets:
+        if isinstance(target, str):
+            atom = atom_db.get_atom(target)
+            if isinstance(atom, Link):
+                new_targets.append(
+                    LinkT(
+                        **{"type": atom.named_type, "targets": parse_targets(atom.targets, atom_db)}
```
angeloprobst, line 26:
```suggestion
                        **{"type": atom.named_type, "targets": parse_targets(atom.targets, atom_db)}
```

src/das_agent/main.py
```diff
@@ -16,21 +18,39 @@ def get_request(node: DASAgentNode):
...
+                    )
```
angeloprobst:
```suggestion
                    LinkT(
                        type=atom.named_type, targets=parse_targets(atom.targets, atom_db)
                    )
```

src/das_agent/main.py
```diff
@@ -16,21 +18,39 @@ def get_request(node: DASAgentNode):
...
+                )
+            else:
+                new_targets.append(NodeT(**{"type": atom.named_type, "name": atom.name}))
```
angeloprobst:
```suggestion
                new_targets.append(NodeT(type=atom.named_type, name=atom.name))
```

src/das_agent/main.py
```diff
@@ -16,21 +18,39 @@ def get_request(node: DASAgentNode):
...
         if target.get("atom_type") == "link":
             new_targets.append(LinkT(**target))
         elif target.get("atom_type") == "node":
             new_targets.append(NodeT(**{"type": target["type"], "name": target["name"]}))
         elif target.get("handle"):
-            new_targets.append(atom_db.get_atom(target["handle"]))
+            atom = atom_db.get_atom(target["handle"])
+            if isinstance(atom, Link):
+                new_targets.append(
+                    LinkT(
```
angeloprobst: Ditto.

src/das_agent/main.py
```diff
@@ -16,21 +18,39 @@ def get_request(node: DASAgentNode):
...
+                        **{"type": atom.named_type, "targets": parse_targets(atom.targets, atom_db)}
+                    )
+                )
+            else:
+                new_targets.append(NodeT(**{"type": atom.named_type, "name": atom.name}))
```
angeloprobst: Ditto.

src/link_creation_agent/equivalence_processor.cc
andre-senna: Ah, now I see the first `}` in line 16. So it seems like the brackets match BUT,  nonetheless, you should fix the `{` and `}` placement to make it clearer.

src/link_creation_agent/implication_processor.cc
```diff
@@ -0,0 +1,155 @@
...
+        {
+        "LINK_TEMPLATE", "Expression", "3",
+            "NODE", "Symbol", "EVALUATION",    
+            "LINK", "Expression", "2",      
+                "NODE", "Symbol", "PREDICATE"},
+                p,
+           {"LINK_TEMPLATE", "Expression", "2",          
+                "NODE", "Symbol", "CONCEPT",       
+                "VARIABLE", "PX"}
+    };
```
andre-senna: Same comment about `{`. and `}` placement.

src/link_creation_agent/implication_processor.cc
```diff
@@ -0,0 +1,155 @@
...
+                        "VARIABLE", "C",
+                "LINK_TEMPLATE", "Expression", "3",
+                    "NODE", "Symbol", "EVALUATION",
+                    "LINK", "Expression", "2",
+                        "NODE", "Symbol", "PREDICATE"},
+                        p2,
+                    {"LINK_TEMPLATE", "Expression", "2",
+                        "NODE", "Symbol", "CONCEPT",
+                        "VARIABLE", "C"}
+    };
```
andre-senna: Same comment about `{`. and `}` placement.

andre-senna (approved):

angeloprobst (approved):

PR #332, opened 2025-03-26
Add Make clause to run tests disregarding test cache

andre-senna:
Any better way to do this? @Pedrobc89  @angeloprobst


andre-senna:
Already implemented by #331

PR #335, opened 2025-03-27
Use 0.0.0.0 as services host.

arturgontijo:
This way we still have things working for linux and also macos Docker containers.

### DIFF
--- a/src/query_engine/query_element/Source.cc
+++ b/src/query_engine/query_element/Source.cc
@@ -11,7 +11,19 @@ Source::Source(const string& attention_broker_address) {
     this->attention_broker_address = attention_broker_address;
 }
 
-Source::Source() : Source("localhost:" + Source::DEFAULT_ATTENTION_BROKER_PORT) {}
+Source::Source() {
+    string attention_broker_address = Utils::get_environment("DAS_ATTENTION_BROKER_ADDRESS");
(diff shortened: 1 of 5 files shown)


src/query_engine/query_element/Source.cc
```diff
@@ -11,7 +11,19 @@ Source::Source(const string& attention_broker_address) {
...
+    string attention_broker_address = Utils::get_environment("DAS_ATTENTION_BROKER_ADDRESS");
+    string attention_broker_port = Utils::get_environment("DAS_ATTENTION_BROKER_PORT");
+    if (attention_broker_address.empty()) {
+        attention_broker_address = "localhost";
+    }
+    if (attention_broker_port.empty()) {
+        attention_broker_address += ":" + DEFAULT_ATTENTION_BROKER_PORT;
+    } else {
+        attention_broker_address += ":" + attention_broker_port;
+    }
```
angeloprobst, line 31: I think this code should be a static method named `get_attention_broker_address` and keep this non-param constructor similar to the original but calling the static method, as follows:
```cpp
Source::Source() : Source(Source::get_attention_broker_address()) {}
```

arturgontijo: Done at [31b3641](https://github.com/singnet/das/pull/335/commits/31b3641a6f72a80e74183ce98e10fce982ea5e6b)

angeloprobst (approved):

Issue #328, opened 2025-03-25
Create an abstract class for AtomDB and implement it for Redis+MongoDB


arturgontijo:
The idea is to create a new lib, called `AtomDB`, with an abstract class that exposes an interface (API) to an inner backend (eg. RAM, Redis+MongoDB, Postgresql, etc).

That abstract class must have attributes and methods used to read/write from/to the inner backend (eg. `query_for_patterns()`, `create_link()`, etc).

Once we have the abstract class we must use it to implement a concrete class that uses Redis+MongoDB as its backend so we can use it in the current `query_engine` (read) and in the link creation code (write).

The goal of this new approach is to have a design where during the instantiation one could set which backend the `AtomDB` must use and that backend must be accessed by the `AtomDB` itself only (eg. create a `DASServer` that instantiates an `AtomDB` running on a Redis+MongoDB backend).

arturgontijo:
Completed by #338 and #347

PR #347, opened 2025-04-01
[#328] Add HandleSet type

arturgontijo:
This PRs introduce a new `atomdb_api_types::HandleSet` abstract class that:
 - Holds multiple `data` via `append(void* data)`.
 - Returns total `data` length/size via `size()`.
 - Returns one element of `data` via `next()` (and a `nullptr` when there is no more elements).

It also implements a concrete class of it, called `HandleSetRedis`, that holds `redisReply*` elements.

This PR also moves Redis+MongoDB types and implementations to `RedisMongoDBAPITypes.h` and `RedisMongoDBAPITypes.cc`, leaving `AtomDBAPITypes.h` with abstracts classes only.

### DIFF
--- a/src/atomdb/AtomDBAPITypes.h
+++ b/src/atomdb/AtomDBAPITypes.h
@@ -24,22 +24,22 @@ namespace atomdb_api_types {
 // -------------------------------------------------------------------------------------------------
 
 class HandleList {
-    public:
-     HandleList() {}
-     virtual ~HandleList() {}
- 
-     virtual const char* get_handle(unsigned int index) = 0;
-     virtual unsigned int size() = 0;
+   public:
+    HandleList() {}
+    virtual ~HandleList() {}
+
+    virtual const char* get_handle(unsigned int index) = 0;
+    virtual unsigned int size() = 0;
 };
 
 class HandleSet {
-    public:
-     HandleSet() {}
-     virtual ~HandleSet() {}
- 
-     virtual unsigned int size() = 0;
-     virtual void append(void* data) = 0;
-     virtual char* next() = 0;
+   public:
+    HandleSet() {}
+    virtual ~HandleSet() {}
+
+    virtual unsigned int size() = 0;
+    virtual void append(void* data) = 0;
+    virtual char* next() = 0;
 };
 
 class AtomDocument {
--- a/src/atomdb/AtomDBAPITypes.h
+++ b/src/atomdb/AtomDBAPITypes.h
@@ -32,14 +32,19 @@ class HandleList {
     virtual unsigned int size() = 0;
 };
 
+class HandleSetIterator {
+    public:
+     virtual char* next() = 0;
+};
+
 class HandleSet {
    public:
     HandleSet() {}
     virtual ~HandleSet() {}
 
     virtual unsigned int size() = 0;
-    virtual void append(void* data) = 0;
-    virtual char* next() = 0;
+    virtual void append(shared_ptr<HandleSet> other) = 0;
+    virtual shared_ptr<HandleSetIterator> get_iterator() = 0;
 };
 
 class AtomDocument {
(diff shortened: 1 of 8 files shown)


src/atomdb/RedisMongoDB.cc
```diff
@@ -141,13 +141,15 @@ void RedisMongoDB::mongodb_setup() {
     }
 }
 
-vector<string> RedisMongoDB::query_for_pattern(std::shared_ptr<char> pattern_handle) {
+shared_ptr<atomdb_api_types::HandleSet> RedisMongoDB::query_for_pattern(
```
andre-senna, line 144: This is not the target line number but Github don't allowed me to add a comment at the proper place because it was not changed in this PR. Anyway...

The env var names to setup AttentionBroker are misleading.

```
DAS_ATTENTION_BROKER_ADDRESS
DAS_ATTENTION_BROKER_PORT
```

Either it should be only a single variable `DAS_ATTENTION_BROKER_ADDRESS` OR two variables suffixed with `_HOSTNAME` and `_PORT`

src/atomdb/AtomDBAPITypes.h
```diff
@@ -42,32 +32,14 @@ class HandleList {
...
-    unsigned int size();
+    HandleSet() {}
+    virtual ~HandleSet() {}
 
-   private:
-    unsigned int handles_size;
-    char** handles;
-    redisReply* redis_reply;
+    virtual unsigned int size() = 0;
+    virtual void append(void* data) = 0;
```
andre-senna: I believe a more elegant solution would be to have:

```
virtual void append(shared_ptr<HandleSet> other)
```

So, inside RedisMongoDBAPITypes.cc you'd do something like:

```
void HandleSetRedis::append(shared_ptr<HandleSet> other) {
    shared_ptr<HandleSet> handle_set_redis = dynamic_cast<HandleSetRedis>(other); 
    for (auto reply : handle_set_redis->replies) {
        this->replies.push_back(reply)
        this->handles_size += reply->elements;    
    }
}
```

The extra overhead here is minimal. It's the creation (and deletion) of one extra HandleSet object per page. And the design benefits are many. Future implementations of HandleSet to other databases would have a more orthogonal behavior.

andre-senna: This is pseudo-code. I didn't tried it ;-)

src/atomdb/RedisMongoDBAPITypes.h
```diff
@@ -0,0 +1,67 @@
...
+
+class HandleSetRedis : public HandleSet {
+   public:
+    HandleSetRedis();
+    HandleSetRedis(redisReply* reply);
+    ~HandleSetRedis();
+
+    unsigned int size();
+    void append(void* data);
+    char* next();
```
andre-senna: I didn't like this design. You mixed both, the container and the iterator in the same class. This is very error prone.

I better approach, in my opinion, would be to have a separate class HandleSetRedisIterator, which would have only one method `next()` which returns the next element or NULL when finished.

HandleSetRedisIterator constructor would be something like:

```
HandleSetRedisIterator(const HandleSetRedis& handle_set)
```

I suggest making HandleSetRedisIterator a friend class of HandleSetRedis:

```
class HandleSetRedis : public HandleSet {
   friend class HandleSetRedisIterator
   public:
    HandleSetRedis();
    [...]
```

This way HandleSetRedisIterator can access private stuff inside HandleSetRedis so you don't need to copy anything. You just need to have the indexes.

So in HandleSetRedisIterator constructor you initialize indexes just like you did but when next() comes to the end, it would return NULL without reseting the indexes. It means the iterator couldn't be reused.

andre-senna (changes requested):

andre-senna (approved):

PR #351, opened 2025-04-02
[#320] Implementation of DAS service bus

### LINKED ISSUE #320: Implement service BUS using BusNode

andre-senna:
Resolves #320 

This is the implementation of the infra-structure required to implement BUS commands, not the BUS commands themselves. I mean we're still missing the elements that actually implement the commands.

A follow-up PR will implement the bus command to pattern matching queries, replacing what's currently in `query_engine/DASNode` see this card: https://github.com/singnet/das/issues/350


src/service_bus/ServiceBus.h
```diff
@@ -18,15 +22,15 @@ namespace service_bus {
  *
  * Elements in the bus can be either command providers or command issuers (or both).
  *
- * Command issuers just use ServiceBus to issue commands. It doesn't provide any service
- * (i.e. it doesn't process any commands) itself. Command providers listen to the bus
- * for commands it's responsible for and process such commands when other bus element
+ * Command issuers just use ServiceBus to issue commands. They doesn't provide any service
+ * (i.e. They doesn't process any commands) itself. Command providers listen to the bus
```
angeloprobst:
```suggestion
 * Command issuers just use ServiceBus to issue commands. They don't provide any service
 * (i.e. They don't process any commands) themselves. Command providers listen to the bus
```

angeloprobst (approved):

Issue #204, opened 2025-01-24
Cache LinkTemplate query results


andre-senna:
LinkTemplate (a piece of the Query Agent) performs queries in AtomDB. The result of such queries can be cached. 

* An important point to keep in mind is that although the query result itself can be cached, the actual query answers require re-priorization by the AttentionBroker.
* LinkTemplates have a handle which can be used as key in the cache

PR #359, opened 2025-04-07
[#204] adding local in-memory caches over Redis/Mongo returns

angeloprobst:
Progress on #204


andre-senna (changes requested):
We already have abstract classes for HandleList and AtomDocument so we shouldn't define the cache explicitly for RedisMongo subtypes. Otherwise we'd need to re-implement the cache whenever we implement new AtomDB based in other DBMS.  neither we should expose the data structures used to store them (e.g. an std::unordered_map) in any AtomDB, otherwise we'd need to replicate changes in every AtomDB whenever we change the way we keep data in the cache.

My suggestion is to have a separate class for the cache. And the methods in this class API should expect/store/return objects of the abstract types. The actual storage data structures should be encapsulated inside this class and once add/get methods should be exposed.

angeloprobst:
> My suggestion is to have a separate class for the cache. And the methods in this class API should expect/store/return objects of the abstract types. The actual storage data structures should be encapsulated inside this class and once add/get methods should be exposed.

@andre-senna 
Done! Pls review it again. Thx.

src/atomdb/AtomDBCache.h
```diff
@@ -0,0 +1,125 @@
...
+     * @brief Destructor.
+     */
+    virtual ~AtomDBCache() {}
+
+    /**
+     * @brief Get the instance of this class.
+     *
+     * @return The instance of this class.
+     */
+    static shared_ptr<AtomDBCache> get_instance();
```
andre-senna: This design of lazy instantiation on the first call to `get_instance()` is dangerous in C++ because the language doesn't guarantee the order in which static variables are initialized. So when you have more than one static variable in the same class (in this class we have the instance and a mutex) you make experiment unexpected behavior.

Typically you can solve this by adding an explicit static method to initialize all static variables which must be called before any calls to `get_instance()`. 

Another design suggestion is to add another class to act like the singleton builder/getter (the caller should call a static method in this second class in order to get the singleton instance). The point of having this is to make easier to write tests since you'd be allowed to instantiate more than one object of the "singleton" class.

Something like:

```
class AtomDBCacheSingleton {
  public: 
    static void initialize();
    static shared_ptr<AtomDB> get_instance();
}

File: src/atomdb/AtomDBCache.cc [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+    if (instance == nullptr) {
+        instance = shared_ptr<AtomDBCache>(new AtomDBCache());
+    }
+    return instance;
+}
+
+shared_ptr<atomdb_api_types::AtomDocument> AtomDBCache::get_atom_document(const char* handle) {
+    lock_guard<mutex> lock(atom_doc_cache_mutex);
+    if (atom_doc_cache.find(handle) != atom_doc_cache.end()) {
+        DEBUG_PRINT("AtomDBCache::get_atom_document :: Found in cache " << handle);
```
andre-senna: Minor comment: the terms commonly used for cache are: "cache hit" and "cache miss".

andre-senna (approved):

src/atomdb/AtomDBCache.cc
```diff
@@ -0,0 +1,71 @@
...
+using namespace std;
+using namespace atomdb;
+
+// #define CACHE_DEBUG
+
+#ifdef CACHE_DEBUG
+#define DEBUG_PRINT(x) cout << x << endl
+#else
+#define DEBUG_PRINT(x)
+#endif
```
eddiebrissow:
```suggestion
#include "Logger.h"
```

src/atomdb/AtomDBCache.cc
eddiebrissow:
```suggestion
        LOG_DEBUG("AtomDBCache::get_atom_document :: Found in cache " << handle);
```

src/atomdb/AtomDBCache.cc
```diff
@@ -0,0 +1,71 @@
...
+    return instance;
+}
+
+shared_ptr<atomdb_api_types::AtomDocument> AtomDBCache::get_atom_document(const char* handle) {
+    lock_guard<mutex> lock(atom_doc_cache_mutex);
+    if (atom_doc_cache.find(handle) != atom_doc_cache.end()) {
+        DEBUG_PRINT("AtomDBCache::get_atom_document :: Found in cache " << handle);
+        return atom_doc_cache[handle];
+    }
+    DEBUG_PRINT("AtomDBCache::get_atom_document :: Not found in cache " << handle);
```
eddiebrissow:
```suggestion
    LOG_DEBUG("AtomDBCache::get_atom_document :: Not found in cache " << handle);
```

src/atomdb/AtomDBCache.cc
```diff
@@ -0,0 +1,71 @@
...
+void AtomDBCache::add_atom_document(const char* handle,
+                                    shared_ptr<atomdb_api_types::AtomDocument> document) {
+    lock_guard<mutex> lock(atom_doc_cache_mutex);
+    atom_doc_cache[handle] = document;
+}
+
+shared_ptr<atomdb_api_types::HandleSet> AtomDBCache::query_for_pattern(const char* pattern_handle) {
+    lock_guard<mutex> lock(pattern_matching_cache_mutex);
+    if (pattern_matching_cache.find(pattern_handle) != pattern_matching_cache.end()) {
+        DEBUG_PRINT("AtomDBCache::query_for_pattern :: Found in cache " << pattern_handle);
```
eddiebrissow:
```suggestion
        LOG_DEBUG("AtomDBCache::query_for_pattern :: Found in cache " << pattern_handle);
```

src/atomdb/AtomDBCache.cc
```diff
@@ -0,0 +1,71 @@
...
+        return pattern_matching_cache[pattern_handle];
+    }
+    DEBUG_PRINT("AtomDBCache::query_for_pattern :: Not found in cache " << pattern_handle);
```
eddiebrissow:
```suggestion
    LOG_DEBUG("AtomDBCache::query_for_pattern :: Not found in cache " << pattern_handle);
```

src/atomdb/AtomDBCache.cc
```diff
@@ -0,0 +1,71 @@
...
+void AtomDBCache::add_pattern_matching(const char* pattern_handle,
+                                       shared_ptr<atomdb_api_types::HandleSet> results) {
+    lock_guard<mutex> lock(pattern_matching_cache_mutex);
+    pattern_matching_cache[pattern_handle] = results;
+}
+
+shared_ptr<atomdb_api_types::HandleList> AtomDBCache::query_for_targets(const char* link_handle) {
+    lock_guard<mutex> lock(handle_list_cache_mutex);
+    if (handle_list_cache.find(link_handle) != handle_list_cache.end()) {
+        DEBUG_PRINT("AtomDBCache::query_for_targets :: Found in cache " << link_handle);
```
eddiebrissow:
```suggestion
        LOG_DEBUG("AtomDBCache::query_for_targets :: Found in cache " << link_handle);
```

src/atomdb/AtomDBCache.cc
```diff
@@ -0,0 +1,71 @@
...
+        return handle_list_cache[link_handle];
+    }
+    DEBUG_PRINT("AtomDBCache::query_for_targets :: Not found in cache " << link_handle);
```
eddiebrissow:
```suggestion
    LOG_DEBUG("AtomDBCache::query_for_targets :: Not found in cache " << link_handle);
```

eddiebrissow (approved):

src/atomdb/AtomDBCache.h
angeloprobst: Done.

src/atomdb/AtomDBCache.cc
angeloprobst: Done.

src/atomdb/AtomDBCache.cc
angeloprobst: Done.

andre-senna (approved):

PR #374, opened 2025-04-18
[#361] Add optional filtering of QueryAnswers which can be considered duplicates

### LINKED ISSUE #361: Query Agent - Add parameter to not return same variables OR/AND
Branch: `master` | `link-creation-processors-support`

When executing an OR query with a variable, duplicate results may be returned if the variable values are the same.
Add a parameter to return only one result when variable values are repeated.

Example:

OR  
LINK...  
VARIABLE P  
Concept "human"  
LINK...  
VARIABLE P  
Concept "monkey"  

This returns:

    (P: "is_animal", "human")

    (P: "is_animal", "monkey")

With the new parameter, only one of them (either "human" or "monkey") should be returned to avoid duplication.

[animals.txt](https://github.com/user-attachments/files/19741160/animals.txt)

andre-senna:
#resolves #361 

The definition of "duplicated" QueryAnswers may be fuzzy. QueryAnswer objects have two components:

* List of handles
* Assignments

These components are, somehow, redundant. The list of handles defines a group of atoms in the atomspace which, together, represent an answer to the query. Assignments defines the variable assignment which leads to this group of atoms.

Therefore we may have a situation where, e.g., two different groups of atoms can be obtained by applying the same variable assignment to the query. the result is two or more QueryAnswers which have the same variable assignment but different list of handles. The other way around is also possible.

For specific algorithms, this duplication may not be desired, thus we're adding optional flags which can be set in proxy's constructor to control the behavior of which types os duplication are allowed and which aren't.


eddiebrissow (approved):
Working great!
Tested with Link Creation Agent

Issue #216, opened 2025-01-27
Implement a distributed evolutionary algorithm to evolve query answers


andre-senna:
The idea is to provide an API in DAS to evolve answers to a query using an evolutionary algorithm.

There are a lot of evolutionary algorithms in the literature (see this [wikipedia topic](https://en.wikipedia.org/wiki/Evolutionary_algorithm)) largely used as optimization techniques for a number of different problems. The algorithm we propose here is also an optimization algorithm that can be used to evolve query answers that best match a given query.

## Why evolving a query answer instead of just processing the query?

Pattern Matching queries may have millions or billions of answers. Sometimes we're interested in just a few particular answers that satisfy a given criteria but in order to find them we'd need to iterate through (potentially) all the query answers in order to find them.

With an optimization algorithm, we use the given criteria as fitness function in a way to guide the search through the space of all possible query answers with the goal of finding answers with high fitness value without the need to iterate through all possibilities.

## How to use users' criteria to select query answers?

The user will pass a query and will provide a function to evaluate a given query answer for this query. But how we can use this fitness function to select query answers?

The more intuitive way of doing this is to process the query, evaluate all its results using the fitness functions and sort them before delivering the iterator to the user. So the user would iterate through query answers from the one with highest fitness value to the one with lowest. But this is bad because we'd need to visit all the query answers which is exactly what we need to avoid.

Hence, the idea is to sort query answers in advance somehow in order to visit only the most promising ones.

DAS already have means to do this: the Attention Broker. It uses atoms' importance value in a given context in order to sort the answers of a query before delivering them to the caller. However, Attention Broker uses solely atoms' importance as sorting criteria and atoms' importance is related to the frequency which each atom have been relevant to previous queries in the same context, it isn't related to any specific fitness measure.

So the idea here is to provide a way to use the "feedback" we implicitly get from user regarding the quality of each query answer (i.e. the passed fitness function) in order to stimulate the importance of atoms in such a way that better answers' components (i.e. atoms) receive importance boosts, making the better answers being more likely of being selected earlier.

## The algorithm

The algorithm we propose here is inspired by [BOA: the Bayesian Optimization Algorithm](https://dl.acm.org/doi/10.5555/2933923.2933973) in the sense that it is an evolutionary algorithm which doesn't actually mix individuals from one generation to the next using genetic operators (e.g. crossover, mutation etc) but, instead, it samples the individuals of the population using probabilistic criteria at the beginning of each new generation.

[...]

PR #375, opened 2025-04-22
[#216] Re-structure agents folder

andre-senna:
WIP towards #216 

In this PR we just create skeleton files for the basic class to provide the "evolution" BUS service.
We also move the query engine to a newly created "agents" folder which will contain all the other DAS agents as well.
(the other agents will be moved to "agents" later; I decided not to move them here because it'd make merging
of current under development code in e.g. LinkCreationAgent almost impossible).


eddiebrissow (approved):
LGTM

Issue #139, opened 2024-11-12
Add graceful_shutdown() API to DistributedAlgorithmNode and remove from concrete subclasses


andre-senna:
Known subclasses with their own shutdown API

* QueryNode (AttentionBroker)
* DASNode (AttentionBroker)

PR #382, opened 2025-04-25
[#139] WIP Implementation of StoppableThread and its use in MessageBroker and query agent.

andre-senna:
WIP towards #139 

Although this is a WIP change, the PR itself is OK to be revised and merged.

Here, we implement an abstract generic interface to "stoppable" elements and a StoppableThread, which is a wrapper for std::threads aimed at guaranteeing that it gets properly stopped, joined and destroyed based in external (in the scope of the thread caller) control.

We also make use of this newly created class in MessageBroker and query agent.

Still missing implement its use in distributed_algorithm_node  package and query elements in general.


angeloprobst (approved):
Tested locally, it worked.

Issue #409, opened 2025-05-28
Implement RemoteDistributedAtomSpace in C++


andre-senna:
### Methods

```
// scope is a value in {REMOTE_ONLY, LOCAL_ONLY, LOCAL_AND_REMOTE}
// All scope parameters below should be defaulted to LOCAL_AND_REMOTE

// handle is a char*
// I'm not sure how to store *targets list*

* const Atom& get_atom(handle, scope)
* const Link& get_link(type, *targets list*, scope)
* const Node& get_node(type, name, scope)
* shared_ptr<PatternMatchingQueryProxy> pattern_matching_query(query, answers_count = 0) // answers_count == 0 indicates ALL answers
* unsigned int pattern_matching_count(query) // query is NULL or empty indicates count ALL atoms in the knowledge base
* void pattern_matching_fetch(query, answers_count = 0) // answers_count == 0 indicates ALL answers
* char*  add_node(type, name) // return handle
* char*  add_link(type, *targets list*) // return handle
* void commit_changes(scope)
```

### State

```
* shared_ptr<AtomDB> db
* shared_ptr<ServiceBus> bus
* // a wrapper for a HandleTrie which maps from handle -> atom 
```

### Methods behavior

```
* get_atom(handle)
* get_link(type, *targets list*)
* get_node(type, name)
```

LOCAL_ONLY and REMOTE_ONLY look for the atom in db or by issuing a service bus command accordingly.
For LOCAL_AND_REMOTE, look for the atom/link/node in the db. If present, return it. Otherwise, issue a command in the bus.

```
* pattern_matching_query(query, answers_count = 0)
* pattern_matching_count(query)
* pattern_matching_fetch(query, answers_count = 0)
```

Queries are issued in the service bus

```
* char*  add_node(type, name)
* char*  add_link(type, *targets list*)
```

Atoms are instantiated inside add_node() and add_link(). Atom handle is computed (and NOT stored in the Atom object). Atom pointer is stored directly in the HandleTrie using the computed handle as key.

```
* void commit_changes()
```

If LOCAL scope is selected (LOCAL_ONLY or LOCAL_AND_REMOTE), commit ALL atoms in local db by sending them to the server (by issuing a command in the service bus). Atoms should be kept in the local db. The command should wait this sync ends before going to the next step.

Once local changes are committed, issue a DB COMMIT command in the service bus.

PR #430, opened 2025-06-03
[#409] AtomSpace client

angeloprobst:
Progress on #409 

Functionality is working, still missing unit tests.


src/atom_space/AtomSpace.h
```diff
@@ -0,0 +1,333 @@
...
+
+    virtual string to_string() override {
+        return "Node(type: " + this->type + ", name: " + this->name + ")";
+    }
+
+    /**
+     * @throws std::runtime_error Always, as merge is not implemented for Node.
+     */
+    void merge(HandleTrie::TrieValue* other) override {
+        throw runtime_error("Merge not implemented for Node");
```
andre-senna: Please implement empty merge() instead of throwing an exception

src/atom_space/AtomSpace.h
```diff
@@ -0,0 +1,333 @@
...
+    bool delete_targets;    ///< Whether to delete target atoms on destruction.
+
+    /**
+     * @brief Construct a Link.
+     * @param type The type of the link.
+     * @param targets The target atoms of the link.
+     * @param delete_targets If true, Link will delete its targets on destruction.
+     * @throws std::runtime_error if type is empty or targets.size() < MINIMUM_TARGETS_SIZE.
+     */
+    Link(const string& type, const vector<Atom*>& targets, bool delete_targets = false)
```
andre-senna: I understand the appeal to have the flag delete_targets but I think it's dangerously error prone. Specially if you pass it in link's constructor instead of when you are actually trying to destroy it. If you think it's really required we should discuss a better design for it. Otherwise, I think we might just remove it..

src/atom_space/AtomSpace.h
```diff
@@ -0,0 +1,333 @@
...
+        : type(type), targets(targets), delete_targets(delete_targets) {
+        if (this->type.empty()) {
+            throw runtime_error("Link type must not be empty");
+        }
+        if (this->targets.size() < MINIMUM_TARGETS_SIZE) {
```
andre-senna: This is wrong. there's not such minimum limit, or at least the limit should be 1. We definitely need "unary" links with only 1 target.

src/atom_space/AtomSpace.h
```diff
@@ -0,0 +1,333 @@
...
+        result.erase(result.size() - 2);  // Remove the last comma and space
+        result += "])";
+        return result;
+    }
+
+    /**
+     * @throws std::runtime_error Always, as merge is not implemented for Link.
+     */
+    void merge(HandleTrie::TrieValue* other) override {
+        throw runtime_error("Merge not implemented for Link");
```
andre-senna: Please implement empty merge() instead of throwing an exception

src/atom_space/AtomSpace.h
```diff
@@ -0,0 +1,333 @@
...
+     * This static utility computes a handle for a Link using the provided type and target atoms,
+     * without requiring an instantiated Link object.
+     *
+     * @param type The type of the link.
+     * @param targets The target atoms of the link.
+     * @return A newly allocated char array containing the handle (caller must free).
+     * @throws std::runtime_error if a target is not a Node or Link.
+     * @note Caller is responsible for freeing the returned handle.
+     */
+    static char* compute_handle(const string& type, const vector<Atom*>& targets) {
```
andre-senna: I'm OK with this method but we should have another compute_handle(type, vector_of_handles).

andre-senna (changes requested):

src/atom_space/AtomSpace.h
angeloprobst: Done.

src/atom_space/AtomSpace.h
angeloprobst: Removed.

src/atom_space/AtomSpace.h
angeloprobst: `MINIMUM_TARGETS_SIZE` changed to 1.

src/atom_space/AtomSpace.h
angeloprobst: Done.

src/atom_space/AtomSpace.h
angeloprobst: Done.

andre-senna (approved):

andre-senna (approved):

PR #507, opened 2025-06-18
Add `atom_space::LinkTemplateInterface`

arturgontijo:


src/atom_space/BUILD
```diff
@@ -30,3 +30,10 @@ cc_library(
         "//hasher:hasher_lib",
     ],
 )
+
+cc_library(
```
andre-senna, line 35: link_template_interface should be added as a dependency of atom_space_lib

src/tests/cpp/BUILD
```diff
@@ -562,6 +562,7 @@ cc_test(
     ],
     linkstatic = 1,
     deps = [
+        "//agents/query_engine:query_engine_lib",
```
andre-senna: Shouldn't this dependency be on //atom_space:link_template_interface?

andre-senna (approved):
Please address the minor comments before merging

PR #518, opened 2025-06-23
[#476] Link Creation Agent BUS implementation 

### LINKED ISSUE #476: Add support for BUS communication to the Link Creation Agent

eddiebrissow:
closes #476


src/service_bus/ServiceBus.cc
```diff
@@ -7,6 +7,7 @@ using namespace service_bus;
 
 string ServiceBus::PATTERN_MATCHING_QUERY = "pattern_matching_query";
 string ServiceBus::QUERY_EVOLUTION = "query_evolution";
+string ServiceBus::LINK_CREATE_REQUEST = "link_create_request";
```
andre-senna: Please follow the naming style of the other already implemented commands. My suggestion is to use "LINK_CREATION"

src/commons/Config.cc
```diff
@@ -0,0 +1,18 @@
+#include "Config.h"
```
andre-senna: I understand why you use a config file but the proper way to achieve this with the BUS design is to let everything you'd add to the config file be a parameter in the proxy. Your config file design, although elegant, should not be used here. You should consider moving it into commons for future use and make all parameters be proxy parameters as they are supposed to be.

src/agents/link_creation_agent/LinkCreationRequestProcessor.cc
```diff
@@ -0,0 +1,55 @@
...
+LinkCreationRequestProcessor::~LinkCreationRequestProcessor() { delete this->link_creation_agent; }
+
+shared_ptr<BusCommandProxy> LinkCreationRequestProcessor::factory_empty_proxy() {
+    shared_ptr<LinkCreationRequestProxy> proxy(new LinkCreationRequestProxy());
+    return proxy;
+}
+
+void LinkCreationRequestProcessor::run_command(shared_ptr<BusCommandProxy> proxy) {
+    auto request_proxy = dynamic_pointer_cast<LinkCreationRequestProxy>(proxy);
+    if (request_proxy->get_command() == LinkCreationRequestProxy::ABORT) {
```
andre-senna: This is wrong. ABORT commands are proxy-level commands, not BUS-level ones. In addition to this, ABORT command is already being implemented in the BaseProxy which we agreed (in a discussion on our Mattermost channel) should be used.

andre-senna (changes requested):

src/tests/cpp/link_creation_agent_test.cc
```diff
@@ -4,51 +4,48 @@
...
+        this->buffer_file_path = "test_buffer.bin";
+        this->metta_file_path = ".";
+        this->save_links_to_metta_file = true;
+        this->save_links_to_db = false;
+        this->server_id = "localhost:7003";
     }
 
     void TearDown() override {
-        remove("test_config.cfg");
+        // remove("test_config.cfg");
```
arturgontijo: I think we can remove this.

arturgontijo:
There are some missing new lines at EOF, remember to run `make format-all` before merging this.
LGTM

arturgontijo (approved):

src/agents/link_creation_agent/LinkCreationRequestProxy.cc
```diff
@@ -0,0 +1,33 @@
...
+    this->command = ServiceBus::LINK_CREATION;
+    this->args.insert(this->args.end(), tokens.begin(), tokens.end());
+}
+
+LinkCreationRequestProxy::~LinkCreationRequestProxy() {}
+
+void LinkCreationRequestProxy::pack_command_line_args() { tokenize(this->args); }
+
+void LinkCreationRequestProxy::set_default_parameters() {
+    this->parameters[LinkCreationRequestProxy::Parameters::QUERY_INTERVAL] = 800;
```
andre-senna: Make an explicit cast to `(unsigned int) 800` otherwise it'll assumed to be `long`.

andre-senna (approved):

PR #519, opened 2025-06-24
[#499] WIP - implementation of population sampling and individuals selection

andre-senna:
WIP towards #499 

The whole evolution dynamics is already working but we're still missing AttentionBroker update based on selected individuals.


src/agents/query_engine/PatternMatchingQueryProcessor.cc
```diff
@@ -80,6 +80,7 @@ void PatternMatchingQueryProcessor::update_attention_broker_single_answer(
             for (unsigned int i = 0; i < query_result_size; i++) {
                 execution_stack.push(string(query_result->get_handle(i)));
             }
+        } else {
```
arturgontijo, line 83: Leftover or should we add a logic here?

arturgontijo (approved):

PR #520, opened 2025-06-24
Implement add and delete logic.

arturgontijo:
This PR implements the following methods of `RedisMongoDB` class:
```
add_node(const atomspace::Node* node)
add_nodes(const vector<atomspace::Node*>& nodes)

add_link(const atomspace::Link* node)
add_links(const vector<atomspace::Link*>& links)

delete_atom(const char* handle)
delete_atoms(const vector<string>& handles)
```
With the `delete` methods we are now able to proper unit test all the methods by adding documents, checking if they exist and then deleting them.

It also creates two new ways to construct a `MongodbDocument` object (from a `Node` and from a `Link`):
```
MongodbDocument(const atomspace::Node* node)
MongodbDocument(const atomspace::Link* link)
```

A following PR will add the Redis logic to the `add_link()` and `add_links()`

### DIFF
--- a/src/atomdb/redis_mongodb/RedisMongoDB.cc
+++ b/src/atomdb/redis_mongodb/RedisMongoDB.cc
@@ -292,7 +292,7 @@ bool RedisMongoDB::link_exists(const char* link_handle) {
             reply->view().find(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS]) != reply->view().end());
 }
 
-vector<string> RedisMongoDB::links_exist(const vector<string>& link_handles) {
+set<string> RedisMongoDB::links_exist(const vector<string>& link_handles) {
     if (link_handles.empty()) return {};
 
     auto conn = this->mongodb_pool->acquire();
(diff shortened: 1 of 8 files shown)


src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -258,77 +308,121 @@ vector<string> RedisMongoDB::links_exist(const vector<string>& link_handles) {
...
     vector<string> existing_links;
-    for (const auto& doc : cursor) {
-        auto values = doc["values"].get_array().value;
-        for (auto&& val : values) {
-            existing_links.push_back(val.get_string().value.data());
+    for (const auto& view : cursor) {
+        auto it = view.find("targets");
+        if (it != view.end() && it->type() == bsoncxx::type::k_array) {
+            auto doc_id = view.find(MONGODB_FIELD_NAME[MONGODB_FIELD::ID]);
+            existing_links.push_back(doc_id->get_string().value.data());
```
andre-senna: I think existing_links should be a set, rather than a vector. Unless you see a specific use-case for it as a vector.

andre-senna (approved):

PR #525, opened 2025-06-27
[AtomDB] Replace char* with string

arturgontijo:
This PR aims to simplify the way callers use `AtomDB` by replacing the `char*` by `std::string` in its methods parameters and returns.

### DIFF
--- a/src/agents/query_engine/PatternMatchingQueryProcessor.cc
+++ b/src/agents/query_engine/PatternMatchingQueryProcessor.cc
@@ -73,7 +73,7 @@ void PatternMatchingQueryProcessor::update_attention_broker_single_answer(
         // Updates joint answer (stimulation)
         joint_answer.insert(handle);
         // Gets targets and stack them
-        auto query_result = db->query_for_targets(handle);
+        shared_ptr<atomdb_api_types::HandleList> query_result = db->query_for_targets(handle);
         if (query_result != NULL) {  // if handle is link
             unsigned int query_result_size = query_result->size();
             for (unsigned int i = 0; i < query_result_size; i++) {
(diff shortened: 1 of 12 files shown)


src/atomdb/AtomDB.h
```diff
@@ -24,13 +24,12 @@ class AtomDB : public HandleDecoder {
 
     virtual shared_ptr<atomdb_api_types::HandleSet> query_for_pattern(
         const LinkTemplateInterface& link_template) = 0;
-    virtual shared_ptr<atomdb_api_types::HandleList> query_for_targets(shared_ptr<char> link_handle) = 0;
```
arturgontijo, line 27: @andre-senna heads up that I removed this and I'm not sure if this would decrease performance, what do you think?

src/agents/query_engine/PatternMatchingQueryProcessor.cc
```diff
@@ -73,8 +73,7 @@ void PatternMatchingQueryProcessor::update_attention_broker_single_answer(
         // Updates joint answer (stimulation)
         joint_answer.insert(handle);
         // Gets targets and stack them
-        shared_ptr<atomdb_api_types::HandleList> query_result =
-            db->query_for_targets((char*) handle.c_str());
+        auto query_result = db->query_for_targets(handle);
```
andre-senna: Please keep the actual query_result type here (auto -> shared_ptr<atomdb_api_types::HandleList>)

src/atomdb/AtomDB.h
andre-senna: I think it's OK

andre-senna (approved):
I made just a minor comment. Please address it before merging.

Issue #521, opened 2025-06-24
Unify design of query elements atoms and the atom_space


andre-senna:
The requirement for the Mork integration is the ability to extract MeTTa expressions (string) out of atoms (specifically LinkTemplate).

PR #530, opened 2025-07-01
[#521] Creates atom classes in commons

andre-senna:
Including wildcards. WIP towards #521 .
Still missing integration with query elements and better tests of Wildcards.


src/tests/cpp/redis_mongodb_test.cc
```diff
@@ -282,12 +295,12 @@ TEST_F(RedisMongoDBTest, AddAndDeleteNode) {
 
     auto node = new Node("Symbol", "\"test-1\"", custom_attributes);
 
-    auto node_handle = Node::compute_handle(node->type, node->name);
+    auto node_handle = node->handle();
 
     // Check if node exists, if so, delete it
-    auto node_document = db->get_atom_document(node_handle);
+    auto node_document = db->get_atom_document(node_handle.c_str());
```
arturgontijo: Nit: No need to use `c_str()` in all the `db` methods as they now accept `string`

arturgontijo (approved):

src/tests/cpp/redis_mongodb_test.cc
andre-senna: fixed

PR #553, opened 2025-07-14
[#544] Atomdb Benchmark

marcocapozzoli:
WIP.

Currently, only tests for **Add Atom**, **Add Atoms**, **Delete Atom**, **Delete Atoms** and **Get Atom** are implemented.


src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -114,7 +115,7 @@ void RedisMongoDB::mongodb_setup() {
     string url = "mongodb://" + user + ":" + password + "@" + address;
 
     try {
-        mongocxx::instance instance;
+        // mongocxx::instance instance;
```
andre-senna: Delete dead code

src/tests/benchmark/atomdb_benchmark.cc
```diff
@@ -0,0 +1,163 @@
...
+
+        auto t0 = Clock::now();
+        db->add_link(link);
+        auto t1 = Clock::now();
+
+        double ms = chrono::duration<double, milli>(t1 - t0).count();
+        local_add_link_latencies.push_back(ms);
+    }
+
+    lock_guard<mutex> lock(global_mutex);
```
andre-senna: Just to make you are aware that it makes no difference where you put this statement , it'll be executed in the beginning of the scope. In this case, in the beginning of the method. I.e., it makes no difference if you put it here or in line 29 (which is a better place to put it, IMO. It's usual to put this kind of semaphore in the beginning of the method.

andre-senna: Anyway... Looking at the code I *think* you don't want to lock the entire method, as you are doing with lock_guard. Perhaps you should consider using a simpler mutex primitive like this:


```
std::mutex global_latencies_mutex;
```

```
    global_latencies_mutex.lock();
    global_latencies["add_node"].insert(global_latencies["add_node"].end(),
                                        local_add_node_latencies.begin(),
                                        local_add_node_latencies.end());
    global_latencies["add_link"].insert(global_latencies["add_link"].end(),
                                        local_add_link_latencies.begin(),
                                        local_add_link_latencies.end());
    global_latencies_mutex.unlock();
```

marcocapozzoli: @andre-senna OK, but what's the difference between

```
lock_guard<mutex> lock(global_mutex);
// code
```
and
```

global_mutex.lock();
// code
global_mutex.unlock();
```

andre-senna: This:

```
void MyClass::my_method() {
    lock_guard<mutex> lock(global_mutex);
    do_some_expensive_thread_safe_stuff();
    do_thread_unsafe_stuff();
}
```

Is equivalent to this:

```
void MyClass::my_method() {
    global_mutex.lock();
    do_some_expensive_thread_safe_stuff();
    do_thread_unsafe_stuff();
    global_mutex.unlock();
}
```

But not to this (which is the right thing to do):

```
void MyClass::my_method() {
    do_some_expensive_thread_safe_stuff();
    global_mutex.lock();
    do_thread_unsafe_stuff();
    global_mutex.unlock();
}
```

In the later, you lock only the portion of the code that actually need to be locked. When you use lock_guard, all the scope where it's inserted will be locked. You could use it this way to achieve the same effect:

```
void MyClass::my_method() {
    do_some_expensive_thread_safe_stuff();
    {
            lock_guard<mutex> lock(global_mutex);
            do_thread_unsafe_stuff();
    }
}
```

But I don't see the point. It's probably a matter of preference. I think the lock/unlock is cleaner. I'd use the inner scope with lock_guard if exception handling were being made in the locked code. lock_guard is safer to use in these situations because it release the lock if an exception is raised in the scope block. But since we'll just abort if an exception occur then readability of lock/unlock approach is better IMO.

marcocapozzoli: Got it! Now I can better understand the difference between the two ways.

src/tests/benchmark/atomdb_benchmark.cc
```diff
@@ -0,0 +1,694 @@
...
+
+#include "AtomDB.h"
+#include "AtomDBCacheSingleton.h"
+#include "Logger.h"
+#include "MorkDB.h"
+#include "RedisMongoDB.h"
+#include "Utils.h"
+#include "atomdb_benchmark_utils.h"
+
+#define LOG_LEVEL INFO_LEVEL
```
andre-senna, line 18: LOG_LEVEL is supposed to be defined BEFORE `#include "Logger.h"

andre-senna (approved):

PR #567, opened 2025-07-16
[Bug Fix] Return empty vector

marcocapozzoli:


src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -646,17 +646,22 @@ vector<string> RedisMongoDB::add_atoms(const vector<atoms::Atom*>& atoms) {
 }
 
 vector<string> RedisMongoDB::add_nodes(const vector<atoms::Node*>& nodes) {
```
arturgontijo, line 648: Nice catch!

Suggestion:
Add this to the first line of the `add_nodes()`:
```
if (nodes.empty()) return {};
```
Let's also add this one to the first line of `add_atoms()`:
```
if (atoms.empty()) return {};
```

marcocapozzoli:
The PR was replaced by this : https://github.com/singnet/das/pull/569

Issue #540, opened 2025-07-04
[RedisMongoDB] Support `Atom::custom_attributes`


arturgontijo:
`RedisMongoDB` must support the `Atom::custom_attributes` field by reading/writing it to the MongoDB.

Special note: [src/atomdb/redis_mongodb/RedisMongoDB.cc#L141-L147](https://github.com/singnet/das/blob/master/src/atomdb/redis_mongodb/RedisMongoDB.cc#L141-L147)

arturgontijo:
Resolved by https://github.com/singnet/das/pull/574

PR #574, opened 2025-07-18
[#540]  [RedisMongoDB] Support Atom::custom_attributes

arturgontijo:


src/atomdb/redis_mongodb/RedisMongoDBAPITypes.cc
```diff
@@ -145,26 +145,64 @@ unsigned int MongodbDocument::get_size(const string& array_key) {
...
+        bsoncxx::builder::basic::document custom_attributes_doc;
+        for (const auto& [key, value] : custom_attributes) {
+            visit(
+                [&](auto&& arg) {
+                    using T = decay_t<decltype(arg)>;
+                    if constexpr (is_same_v<T, string>) {
+                        custom_attributes_doc.append(bsoncxx::builder::basic::kvp(key, arg));
+                    } else if constexpr (is_same_v<T, unsigned int>) {
+                        custom_attributes_doc.append(
+                            bsoncxx::builder::basic::kvp(key, static_cast<int64_t>(arg)));
```
andre-senna: If mongoldb doesn't support the difference between long and unsigned int we should rather raise an exception here instead of silently storing a unsigned int as long.

The problem with this silent approach is that user's code will break after they get an atom from AtomDB and try to get the value of the unsigned int property because they will do something like get<unsigned int>(key). It's (by far) preferable that we raise the exception when the user try to insert the atom in AtomDB than letting him do it just to raise the exception latter. Note that a user could populate an. atom DB to be used by someone else.

andre-senna (approved):

PR #597, opened 2025-07-25
[#592] Miscelaneous minor fixes

### LINKED ISSUE #592: Minor fixes after integration of Evolution agent

andre-senna:
Resolves #592

### DIFF
--- a/src/agents/evolution/QueryEvolutionProcessor.cc
+++ b/src/agents/evolution/QueryEvolutionProcessor.cc
@@ -83,14 +83,13 @@ void QueryEvolutionProcessor::sample_population(
     vector<string> answer_bundle_vector;
     unsigned int population_size =
         proxy->parameters.get<unsigned int>(QueryEvolutionProxy::POPULATION_SIZE);
-    auto pm = atom_space.pattern_matching_query(
-        proxy->get_query_tokens(),
-        population_size,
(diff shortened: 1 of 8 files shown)


src/agents/evolution/QueryEvolutionProcessor.cc
```diff
@@ -222,14 +228,23 @@ void QueryEvolutionProcessor::correlate_similar(shared_ptr<QueryEvolutionProxy>
                 return;
             }
             token = original_tokens[cursor++];
-            string value = correlation_query_answer->assignment.get(token);
+            if (token == "sentence3") {
```
arturgontijo, line 231: Do we want this here?

andre-senna: Actually no. But it's the better we can do for now. I'm still trying to figure out how to setup correlation query in order for it to work (i.e. in order for it to make evolution as a whole work) so it's pointless to refactor the command API (which is what's is supposed to be required to fix this code).

andre-senna: This is a hard-wired attempt to make the current query in word-query-evolution client work. I just shared this to make sure we both are running evolution on the same code.

arturgontijo (approved):

PR #608, opened 2025-07-28
Fitness function

eddiebrissow:
* Added Fitness Function to multiply strength
* Fixed inference client
* Added temporary method to add inference index patterns
* Changed Evolution Query to start at level 2

Obs: Commented methods/lines will be addressed in a new PR


src/agents/evolution/fitness_functions/FitnessFunctionRegistry.cc
```diff
@@ -6,6 +6,7 @@
 // -------------------------------------------------------------------------------------------------
 // Add your header here
 #include "CountLetterFunction.h"
+#include "StrengthFunction.h"
```
andre-senna: This name is meaningless. Perhaps `MultiplyStrengthFunction`?

src/agents/evolution/fitness_functions/StrengthFunction.cc
```diff
@@ -0,0 +1,30 @@
...
+
+string StrengthFunction::VARIABLE_NAME = "strength";
+
+StrengthFunction::StrengthFunction() { db = AtomDBSingleton::get_instance(); }
+
+float StrengthFunction::eval(shared_ptr<QueryAnswer> query_answer) {
+    float strength = 1.0;
+
+    LOG_INFO("Evaluating strength for " << query_answer->to_string());
+    LOG_INFO("Evaluating strength for " << query_answer->handles.size() << " handles.");
```
andre-senna: `handles` is already printed in the INFO message above. If you really want to see the size explicitly you could use a single combined message to print both.

andre-senna (approved):

PR #617, opened 2025-07-30
Enable support for running multiple benchmarks

marcocapozzoli:
- Makes the benchmark script more generic and adaptable to multiple types (atomdb, pattern_matching_query, etc).
- Facilitates future extensions and simplifies usage via CLI.


src/scripts/run_benchmark.sh
```diff
@@ -90,51 +113,46 @@ parse_args()  {
...
 
     case "$db_size" in
     empty)
-        SENTENCES=100
-        WORD_COUNT=3
-        WORD_LENGTH=3
+        SENTENCES=10
+        WORD_COUNT=1
         ;;
     small)
```
andre-senna, line 124: I'd suggest:

small:
        SENTENCES=10000
        WORD_COUNT=5
        WORD_LENGTH=3
medium:
        SENTENCES=1000000
        WORD_COUNT=5
        WORD_LENGTH=3
large:
        SENTENCES=10000000
        WORD_COUNT=5
        WORD_LENGTH=3
xlarge:
        SENTENCES=100000000
        WORD_COUNT=5
        WORD_LENGTH=3

src/scripts/run_benchmark.sh
```diff
@@ -90,51 +113,46 @@ parse_args()  {
...
+        WORD_COUNT=10000
         ;;
     *)
         echo "Invalid --db value: $db_size" >&2
         exit 1
         ;;
     esac
 
+    # Smaller vocabulary, greater connectivity (V = A^L)
     case "$rel" in
```
andre-senna, line 143: I'd suggest

loosely
        ALPHABET_RANGE="0-25"
        WORD_LENGTH=3
        WORD_COUNT=5
tightly
        ALPHABET_RANGE="0-5"
        WORD_LENGTH=3
        WORD_COUNT=10

src/tests/benchmark/atomdb/atomdb_utils.h
```diff
@@ -58,5 +59,6 @@ double measure_execution_time(Func&& func) {
     auto start = chrono::steady_clock::now();
     func();
     auto end = chrono::steady_clock::now();
-    return chrono::duration<double, milli>(end - start).count();
+    double duration_millis = chrono::duration<double, milli>(end - start).count();
+    return round(duration_millis * 100.0) / 100.0;  // Round to 2 decimal places
```
andre-senna: You shouldn't cut precision when you are measuring or manipulating numbers in general. You only do such simplification when you are about to report to a text file or a GUI component. Internally, you ALWAYS keep maximum precision. If cut precision too early, like you did here, the rounding error will sum up whenever you use this number downstream.

When we integrate to the DB. You put the whole precision in the DB as well. Let whoever is showing these numbers to the final caller to cut down precision, if necessary.

andre-senna (approved):

PR #653, opened 2025-08-08
[#649] Implements a MeTTa parser on top of a lexer previously implemented in #652 

andre-senna:
This is WIP towards #649

We are still missing a concrete class for ParserActions and its actual integration to build LinkSchema when a LinkTemplate is built. This will be implemented in a forthcomming PR.

### DIFF
--- a/src/tests/cpp/metta_parser_test.cc
+++ b/src/tests/cpp/metta_parser_test.cc
@@ -28,8 +28,8 @@ static void check_tokens(const string& tag,
     if (expected_tokens.size() == tokens->size()) {
         for (unsigned int i = 0; i < tokens->size(); i++) {
             LOG_INFO(token_name[expected_tokens[i].first] + "<" + expected_tokens[i].second + "> " +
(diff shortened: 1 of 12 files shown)


src/tests/cpp/metta_parser_test.cc
```diff
@@ -48,6 +80,30 @@ static void check_string(const string& tag,
...
 
+static void check_parse(const string& tag,
+                        const string& metta_str,
+                        bool syntax_ok,
+                        const vector<string>& expected) {
+    LOG_INFO("--------------------------------------------------------------------------------");
+    LOG_INFO("Test case " + tag);
+
+    shared_ptr<TestActions> broker = make_shared<TestActions>();
+    ;
```
arturgontijo, line 91: nit: remove

arturgontijo (approved):

src/tests/cpp/metta_parser_test.cc
andre-senna: Fixed in the next PR

PR #660, opened 2025-08-13
Add consolidated benchmark report in .txt

marcocapozzoli:


src/scripts/run_benchmark.sh
```diff
@@ -341,7 +341,7 @@ scenario_data() {
 
 consolidate_reports() {
     local benchmark=$1
-    python3 ./src/scripts/python/consolidate_benchmark.py "/tmp/${benchmark}_benchmark/${TIMESTAMP}" --scenario "$(scenario_data)" --type "$benchmark"
+    python3 ./src/scripts/python/consolidate_benchmark.py "/tmp/${benchmark}_benchmark/${TIMESTAMP}" --scenario "$(scenario_data)" --type "$benchmark" --report-file "1" --header-text "$(header_to_report)" --output-dir "/home/$USER/.cache/das"
```
andre-senna: Either use only `--report-file` to indicate that the consolidated report file is supposed to be generated OR use `--report-file /path/to/file.txt` to indicate it (the absence of the parameter would indicate to not generate the consolidated report).

The second option is better IMO because you could get rid of `--output-dir` as well.

andre-senna (approved):

PR #671, opened 2025-08-15
[#670] Benchmark testing for the Query Agent

### LINKED ISSUE #670: Create some test scenarios to benchmark Query Agent
Simplequery, Complexquery and Positiveimportance

marcocapozzoli:
Resolves #670


src/agents/query_engine/PatternMatchingQueryProcessor.cc
```diff
@@ -172,6 +174,7 @@ void PatternMatchingQueryProcessor::process_query_answers(
 void PatternMatchingQueryProcessor::thread_process_one_query(
     shared_ptr<StoppableThread> monitor, shared_ptr<PatternMatchingQueryProxy> proxy) {
     STOP_WATCH_START(query_thread);
+    STOP_WATCH_START(benchmark_query_thread);
```
andre-senna, line 177: Why do you need a new stop watch? You are starting and stopping it in the same places as the other one so why not just using the one that was there?

src/agents/query_engine/PatternMatchingQueryProcessor.cc
```diff
@@ -1,7 +1,9 @@
 #include "PatternMatchingQueryProcessor.h"
 // clang-format off
 
+#ifndef LOG_LEVEL
```
andre-senna, line 4: This shouldn't be necessary. Why did you need to put this?

andre-senna: The proper way to solve conflicts in LOG_LEVEL is by moving the define to above the "include" clause which is causing the conflict (usually, a templated class whose code, which uses LOG_LEVEL, is defined in the .h and hence the definition is imported) and disable c-lang format to prevent it from reordering the include clauses. This has already been done here so I don't see why it was causing conflicts in your case.

Anyway, if a conflict really exists, it's better to have it reported as warning in the compiler. If you do as you did, you just ignore this definition of LOG_LEVEL silently.

src/scripts/bazel_build.sh
```diff
@@ -64,7 +64,7 @@ if [ "$BUILD_WHEELS" = true ]; then
     $BAZELISK_RUN_CMD //deps:requirements_python_client.update
 fi
 
-$BAZELISK_BUILD_CMD $BUILD_TARGETS
+$BAZELISK_BUILD_CMD $BUILD_TARGETS "$@"
```
andre-senna, line 67: @levisingularity  @marcocapozzoli 

What's the effect of this change? Levi, please confirm this is OK.

src/scripts/build.sh
```diff
@@ -42,7 +42,7 @@ docker run --rm \
   --volume $LOCAL_WORKDIR:$CONTAINER_WORKDIR \
   --workdir $CONTAINER_WORKSPACE_DIR \
   ${IMAGE_NAME} \
-  ./scripts/bazel_build.sh
+  ./scripts/bazel_build.sh "$@"
```
andre-senna, line 45: The same here @levisingularity  @marcocapozzoli

andre-senna (changes requested):

src/agents/query_engine/PatternMatchingQueryProcessor.cc
marcocapozzoli: When I used `STOP_WATCH_FINISH(benchmark_query_thread, "Benchmark::PatternMatchingQuery");`, I passed "Benchmakr::". The idea was that all benchmark-related logs would have this pattern at the beginning of the "Benchmark::" message. This would make it easier to parset later, as I could filter by it.

src/agents/query_engine/PatternMatchingQueryProcessor.cc
marcocapozzoli: The problem wasn't a conflict. It's this: I need to capture the times generated by STOP_WATCH_FINISH, and it only fires if the log level is DEBUG_LEVEL. I didn't want to set DEBUG_LEVEL directly in the .cc file, as that would "pollute" the logs. So, the easiest solution I found was to leave the #ifndef there and pass DEBUG_LEVEL as an argument in the build command for the binaries I use for testing.

`./src/scripts/build.sh --copt=-DLOG_LEVEL=DEBUG_LEVEL`

This way I can keep it only at DEBUG level in testing.

src/scripts/bazel_build.sh
marcocapozzoli: @andre-senna the answer above explains why I put that there

src/scripts/build.sh
marcocapozzoli: @andre-senna the answer above explains why I put that there

src/agents/query_engine/PatternMatchingQueryProcessor.cc
andre-senna: OK

src/agents/query_engine/PatternMatchingQueryProcessor.cc
andre-senna: Got it. OK

src/scripts/bazel_build.sh
levisingularity: @andre-senna runs the command defined in `$BAZELISK_BUILD_CMD` on the targets in `$BUILD_TARGETS`, forwarding all arguments passed to the script (`"$@"`).

andre-senna: OK. But here I need to double check with @levisingularity  if the change is harmless regarding our building process, pipeline, packaging etc.

src/scripts/build.sh
levisingularity: @andre-senna  forwarding all arguments passed to the script (`"$@"`).

levisingularity (approved):

andre-senna (approved):

src/scripts/bazel_build.sh
levisingularity: Since the parameter is optional, it won’t affect the CI process.

src/scripts/build.sh
levisingularity: Since the parameter is optional, it won’t affect the CI process.

PR #701, opened 2025-08-22
[#644] Use DBMS to gather benchmark test results

levisingularity:
Closes https://github.com/singnet/das/issues/644

PR #705, opened 2025-08-25
[#HOTFIX] Automatically generate .env file during Query Agent initialization in init_environment script

marcocapozzoli:


marcocapozzoli:
There is another PR that is already applying this fix: https://github.com/singnet/das/pull/701

Issue #632, opened 2025-08-04
Fine tune Inference agent parameters in order to solve a toy sentence problem


andre-senna:

eddiebrissow:
Waiting next iteration

andre-senna:
Related to #1009

PR #733, opened 2025-09-03
[#632] Added cache to not process already processed Query Answers

eddiebrissow:
links to #632

### DIFF
--- a/src/agents/link_creation_agent/LinkCreationService.cc
+++ b/src/agents/link_creation_agent/LinkCreationService.cc
@@ -43,7 +43,7 @@ LinkCreationService::~LinkCreationService() {
 void LinkCreationService::save_cache() {
     ofstream file(metta_file_path + "/" + "cache");
     if (file.is_open()) {
-        // NOTE to Reviewer: leaving this code commented out for now, not sure if it's a bug or 
+        // NOTE to Reviewer: leaving this code commented out for now, not sure if it's a bug or
         // I'm not using correctly, but this approach only saves one key and not all the keys.
         //
         // answer_cache->traverse(
--- a/src/agents/link_creation_agent/LinkCreationService.cc
+++ b/src/agents/link_creation_agent/LinkCreationService.cc
@@ -3,9 +3,9 @@
 #include <fstream>
 
 #include "AtomDBSingleton.h"
+#include "Hasher.h"
 #include "Logger.h"
 #include "Utils.h"
-#include "Hasher.h"
 
 using namespace link_creation_agent;
 using namespace std;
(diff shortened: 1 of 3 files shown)


src/agents/link_creation_agent/LinkCreationService.cc
```diff
@@ -35,6 +37,107 @@ LinkCreationService::~LinkCreationService() {
...
+    delete answer_cache;
+}
+
+void LinkCreationService::save_cache() {
+    ofstream file(metta_file_path + "/" + "cache");
+    if (file.is_open()) {
+        // NOTE to Reviewer: leaving this code commented out for now, not sure if it's a bug or
+        // I'm not using correctly, but this approach only saves one key and not all the keys.
+        //
+        // answer_cache->traverse(
```
andre-senna: This approach is what you are supposed to use. Please change save_cache() to use it. The problem with your implementation is the `return true` statement in line 62. When the traverse lambda function return `true`, it means the traversal process should stop. Just change it to return `true`  and all the elements will be visited.

andre-senna: After fixing this, I think you should remove answer_cache_keys as it shouldn't be required to do anything. If you still think you need it, let us discuss the matter in the channel.

src/agents/link_creation_agent/LinkCreationService.cc
```diff
@@ -35,6 +37,107 @@ LinkCreationService::~LinkCreationService() {
...
+        LOG_ERROR("Error opening file: " << this->metta_file_path + "/" + "cache");
+    }
+}
+
+string LinkCreationService::query_answer_hash(shared_ptr<QueryAnswer> query_answer) {
+    vector<string> variables;
+    for (auto& var : query_answer->assignment.table) {
+        variables.push_back(var.second);
+    }
+    auto key = Utils::join(variables, ',');
```
andre-senna: Use `return Hasher::composite_handle(variables)` instead.

andre-senna: But anyway this method doesn't seem correct. You have no guarantee about the order in which elements will be iterated on `assignment.table` so, for instance, if you have these assignments: `{("v1", "h1"), ("v2", "h2")}` the final values in `variable` could be either `["h1", "h2"]` or `["h2", "h1"]`. I'm not sure this is what you want, but I believe not.

Please take a careful look at your code in order to confirm what I say next but I think you can use the `assignment.handles` vector directly. If so, you could remove this entire method because the QueryAnswer handle could be computed by `Hasher::composite_handle(query_answer->handles)`. 

If you find out you really can't use `assignment.handles` directly, you should build a vector like this: `vector = ["v1", "h1", "v2", "h2"]` and return `Hasher::composite_handle(vector)`.

If you need to do it this way, you must remember to SORT the keys before building the vector otherwise you will get the same problem you have with your current implementation.

andre-senna (changes requested):

andre-senna (approved):

PR #775, opened 2025-09-25
Add `ContextBroker` processor and proxy

arturgontijo:
This PR adds the `ContextBroker` element to the DAS architecture.

It is responsible to create, store and restore `context` from `AttentionBroker`.

In order to start the broker, run the following command:
```
Usage: ./bin/context_broker <hostname:port> <start_port:end_port> BUS_IP:PORT [AB_ip:AB_port]

# Example:
./bin/context_broker 0.0.0.0:37700 52000:52999 0.0.0.0:35700 0.0.0.0:37007
```

Once running it will handle all messages from command `ServiceBus::CONTEXT_MANAGER` (`"context_manager"`)

This PR also modifies `src/main/word_query_evolution_main.cc` to use `ContextBrokerProxy`


src/main/context_broker_main.cc
```diff
@@ -0,0 +1,56 @@
...
+    if (argc < 3) {
+        cerr << "Context Broker Server" << endl;
+        cerr << "Usage: " << argv[0]
+             << " <hostname:port> <start_port:end_port> BUS_IP:PORT [AB_ip:AB_port]" << endl;
+        exit(1);
+    }
+
+    std::string server_id = string(argv[1]);
+
+    if (argc == 5) {
```
andre-senna: Make AB parameter mandatory.

src/agents/context_broker/ContextBrokerProcessor.cc
```diff
@@ -0,0 +1,253 @@
...
+using namespace service_bus;
+using namespace attention_broker;
+
+namespace fs = std::filesystem;
+
+// -------------------------------------------------------------------------------------------------
+// Constructors and destructors
+
+ContextBrokerProcessor::ContextBrokerProcessor() : BusCommandProcessor({ServiceBus::CONTEXT_MANAGER}) {
+    AttentionBrokerClient::health_check(true);
```
andre-senna: Add a log message like:

```
if (AttentionBrokerClient::health_check(true)) {
    LOG_INFO("Connected to AttentionBroker at " + AttentionBrokerClient::SERVER_ADDRESS);
}
```

I'll add the same message in other agents that make this check.

src/main/word_query_evolution_main.cc
```diff
@@ -249,10 +252,43 @@ void run(const string& client_id,
...
+    service_bus->issue_bus_command(context_proxy);
+
+    // Wait for ContextBrokerProxy to finish context creation
+    LOG_INFO("Waiting ContextBrokerProxy to finish context creation...");
+    while (!context_proxy->finished()) {
+        Utils::sleep();
+    }
+
+    string context_str = context_proxy->get_key();
+    LOG_INFO("Context " + context_str + " is ready (using ContextBrokerProxy)");
```
andre-senna:
```suggestion
    LOG_INFO("Context " + context_str + " is ready");
```

src/main/word_query_evolution_main.cc
```diff
@@ -249,10 +252,43 @@ void run(const string& client_id,
...
-        context_tag, context1, {{contains_link, sentence_link}, {sentence_link, word_link}}, {});
-    string context_str = context->get_key();
-    LOG_INFO("Context " + context_str + " is ready");
+    // auto context = atom_space.create_context(
+    //     context_tag, context1, {{contains_link, sentence_link}, {sentence_link, word_link}}, {});
+    // string context_str = context->get_key();
+    // LOG_INFO("Context " + context_str + " is ready");
+
+    // ---------------------------------------------------------------------------------------------
+    // Context creation using ContextBrokerProxy (NEW APPROACH)
```
andre-senna:
```suggestion
    // Context creation using ContextBrokerProxy
```

src/agents/context_broker/ContextBrokerProxy.h
```diff
@@ -0,0 +1,179 @@
...
+class ContextBrokerProxy : public BaseQueryProxy {
+   public:
+    // ---------------------------------------------------------------------------------------------
+    // Constructors, destructors and static state
+
+    /**
+     * Empty constructor typically used on server side.
+     * @param use_cache Whether to use cache
+     */
+    ContextBrokerProxy(bool use_cache = true);
```
andre-senna: Constructor called in server side should have no parameters. The idea is that all parameters are passed as command arguments.

src/agents/context_broker/ContextBrokerProxy.h
```diff
@@ -0,0 +1,179 @@
...
+     * @param use_cache Whether to use cache
+     * @param rent_rate AttentionBroker rent rate
+     * @param spreading_rate_lowerbound AttentionBroker lowerbound spreading rate
+     * @param spreading_rate_upperbound AttentionBroker upperbound spreading rate
+     */
+    ContextBrokerProxy(const string& name,
+                       const vector<string>& query,
+                       const vector<pair<QueryAnswerElement, QueryAnswerElement>>& determiner_schema,
+                       const vector<QueryAnswerElement>& stimulus_schema,
+                       bool use_cache,
```
andre-senna: We'll discuss about the last 3 parameters in call. regarding `use_cache`, it should be defaulted to `true` in the constructor called on client side. Because it's defaulted, it could be a property instead of a class field. This is not actually mandatory but it's handy because properties are already tokenized/untokenized in the base class so you don't need to worry about tokenizing/untokenizing it in this class.

src/agents/context_broker/ContextBrokerProxy.h
```diff
@@ -0,0 +1,179 @@
...
+   private:
+    void init(const string& name, bool use_cache);
+
+    // Context-specific members
+    mutex api_mutex;
+    string name;
+    string key;
+    string cache_file_name;
+    string atom_handle;
+    bool use_cache;
```
andre-senna: Remove this (see my comment above)

src/agents/context_broker/ContextBrokerProxy.cc
```diff
@@ -0,0 +1,224 @@
...
+    float spreading_rate_lowerbound,
+    float spreading_rate_upperbound)
+    : BaseQueryProxy(query, name) {
+    // Constructor for query-based context
+    this->command = ServiceBus::CONTEXT_MANAGER;
+    this->determiner_schema = determiner_schema;
+    this->stimulus_schema = stimulus_schema;
+    this->rent_rate = rent_rate;
+    this->spreading_rate_lowerbound = spreading_rate_lowerbound;
+    this->spreading_rate_upperbound = spreading_rate_upperbound;
```
andre-senna: this->atom_handle should be initialized to ""

src/main/context_broker_main.cc
arturgontijo: I did it that way to keep the same pattern we use in other main files (eg `query_engine_main`).
If one does not provide it, it will default to `DEFAULT_ATTENTION_BROKER_ADDRESS` (`localhost:40001`).

src/agents/context_broker/ContextBrokerProxy.cc
arturgontijo: This variable/parameter logic will not be shipped with this PR anymore

src/agents/context_broker/ContextBrokerProcessor.cc
```diff
@@ -0,0 +1,286 @@
...
+    try {
+        if (proxy->args.size() < 2) {
+            Utils::error("Syntax error in query command. Missing implicit parameters.");
+        }
+
+        proxy->untokenize(proxy->args);
+
+        // Do not allow use_cache == false and enforce_cache_recreation == true
+        if (!proxy->get_use_cache() && proxy->get_enforce_cache_recreation()) {
+            LOG_ERROR("use_cache == false and enforce_cache_recreation == true is not allowed");
```
andre-senna: Instead of throwing the error directly, call `Utils::error("Error message", false);` The second (optional) parameter can be used to prevent an exception to be raised.

src/agents/context_broker/ContextBrokerProxy.cc
```diff
@@ -0,0 +1,267 @@
...
+// ---------------------------------------------------------------------------------------------
+// Virtual superclass API and the piggyback methods called by it
+
+bool ContextBrokerProxy::from_remote_peer(const string& command, const vector<string>& args) {
+    LOG_DEBUG("Proxy command: <" << command << "> from " << this->peer_id() << " received in "
+                                 << this->my_id());
+    if (BaseQueryProxy::from_remote_peer(command, args)) {
+        return true;
+    } else if (command == ATTENTION_BROKER_SET_PARAMETERS) {
+        this->rent_rate = std::stod(args[0]);
```
andre-senna: Use `Utils::string_to_float()` instead to make sure an exception is thrown on parsing errors.

andre-senna (approved):

src/main/context_broker_main.cc
andre-senna: We should do the other way around. In QA it's not defaulted anymore. Having this parameter defaulted is error prone.

src/agents/context_broker/ContextBrokerProcessor.cc
arturgontijo: But in this case we do want to raise an exception, right? So it would be just `Utils::error("Error message")`?

andre-senna (approved):

PR #814, opened 2025-10-23
[#807] Add serialization/deserialization methods to AttentionBroker API

### LINKED ISSUE #807: Add a method in AttentionBroker RPC API to load context info from file

andre-senna:
Resolves #807 

AttentionBrokerClient now have two new methods:

```
    static void save_context(const string& context, const string& file_name);
    static void drop_and_load_context(const string& context, const string& file_name);
```

They allow persistence of contexts in disk.

### DIFF
--- a/src/tests/cpp/BUILD
+++ b/src/tests/cpp/BUILD
@@ -435,7 +435,7 @@ cc_test(
 
 cc_test(
     name = "service_bus_test",
-    size = "small",
+    size = "medium",
     srcs = ["service_bus_test.cc"],
     copts = [
         "-Iexternal/gtest/googletest/include",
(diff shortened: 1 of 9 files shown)


eddiebrissow (approved):

src/tests/cpp/BUILD
```diff
@@ -435,7 +435,7 @@ cc_test(
 
 cc_test(
     name = "service_bus_test",
-    size = "small",
+    size = "medium",
```
arturgontijo, line 438: Leftover?

arturgontijo (approved):

src/tests/cpp/BUILD
andre-senna: Actually not. This test is taking 40+ seconds in my workstation. Once in a while it breaks with a timeout. I think we better bump it to medium.

PR #836, opened 2025-10-27
[#824] Added Proxy parameters and processor tests

eddiebrissow:


andre-senna (approved):

src/agents/link_creation_agent/LinkCreationRequestProxy.cc
```diff
@@ -4,17 +4,21 @@
 
 using namespace link_creation_agent;
 
-const string LinkCreationRequestProxy::Parameters::QUERY_INTERVAL =
-    "link_creation_agent.requests_interval";
-const string LinkCreationRequestProxy::Parameters::QUERY_TIMEOUT = "link_creation_agent.query_timeout";
-
-const string LinkCreationRequestProxy::Parameters::ABORT_FLAG = "link_creation_agent.abort_flag";
+const string LinkCreationRequestProxy::MAX_RESULTS = "lca.mr";
```
andre-senna: Use human-readable strings here to make easier integration in the MeTTa interpreter. Something like:

```
const string LinkCreationRequestProxy::MAX_RESULTS = "max_results";
```

andre-senna (changes requested):

src/agents/inference_agent/InferenceProxy.h
```diff
@@ -8,15 +8,18 @@ namespace inference_agent {
...
     static string INFERENCE_REQUEST_TIMEOUT;
-    static string UPDATE_ATTENTION_BROKER_FLAG;  // Interval for inference requests
-    static string REPEAT_REQUEST_NUMBER;         // Whether to repeat the request
-    static string MAX_QUERY_ANSWERS_TO_PROCESS;  // Max number of query answers to process
-    static string RUN_FULL_EVALUATION_QUERY;
+    /** Repeats the inference process N times */
+    static string REPEAT_COUNT;
+    // Control LCA process
+    /** Maximum number of answers to return for each query and LCA queries */
+    static string MAX_RESULTS;
```
andre-senna: This parameter is exactly `BaseQueryProxy::MAX_ANSWERS`. Why are you creating a new parameter instead of re-using the base class?

src/agents/inference_agent/InferenceProxy.h
```diff
@@ -8,15 +8,18 @@ namespace inference_agent {
...
+    /** Controls if LCA should update the attention broker for each query */
+    static string UPDATE_ATTENTION_BROKER;
```
andre-senna: Same here. This is `BaseQueryProxy::ATTENTION_UPDATE_FLAG`

src/agents/link_creation_agent/LinkCreationRequestProxy.h
```diff
@@ -24,13 +24,22 @@ class LinkCreationRequestProxy : public BaseProxy {
    public:
     // ---------------------------------------------------------------------------------------------
 
+    /** Maximum number of answers to return for each query */
```
andre-senna, line 27: Why LinkCreationRequestProxy is a `BaseProxy` and not a `BaseQueryProxy`? 

MAX_RESULTS, UPDATE_ATTENTION_BROKER, and IMPORTANCE_FLAG should be reused from base class.

INFINITE_REQUEST shouldn't exist at all. Having redundant parameters is error prone. What is supposed to happen if the caller passes  INFINITE_REQUEST=true and REPEAT_COUNT=10? My suggestion is to use REPEAT_COUNT=0 to mean infinite requests.

CONTEXT shouldn't be a parameter. It's a mandatory argument os BaseQueryProxy

src/agents/inference_agent/InferenceProxy.h
```diff
@@ -8,15 +8,18 @@ namespace inference_agent {
...
+    /** Run full evaluation query */
+    static string RUN_FULL_EVALUATION_QUERY;  // TODO remove
```
andre-senna: We won't integrate Inference in HE for v1.0 but we'll do it soon. What is required in order to remove this parameter? Can it be just set to a default hardwired value in HE?

andre-senna (changes requested):

src/agents/link_creation_agent/LinkCreationRequestProxy.h
eddiebrissow: It's a `BaseProxy` because LCA doesn't return answers, I'll remove the misleading parameter

andre-senna (approved):

PR #838, opened 2025-10-28
[#795] update python client for new bus elements [Part 2]

marcocapozzoli:
- Add atoms: Atom, Node and Link
- Add Hasher
- Fix unit tests


3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
```diff
@@ -0,0 +1,118 @@
...
+
+        num_property_tokens = int(tokens[0])
+        if num_property_tokens > 0:
+            properties_tokens = tokens[1: 1 + num_property_tokens]
+            self.custom_attributes.untokenize(properties_tokens)
+            del tokens[: 1 + num_property_tokens]
+        else:
+            del tokens[0]
+
+    def log_and_raise_error(self, msg: str) -> None:
```
andre-senna: This should be implemented in a utility class, not here. Other classes elsewhere in the Python library will need to use it. I suggest replicating the Utils::error() (no need to replicate the entire Utils class)

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
```diff
@@ -0,0 +1,118 @@
...
+            and self.custom_attributes.P == other.custom_attributes.P
+        )
+
+    def __ne__(self, other: 'Atom') -> bool:
+        return not self.__eq__(other)
+
+    def copy_from(self, other: 'Atom') -> 'Atom':
+        self.type = other.type
+        self.is_toplevel = other.is_toplevel
+        self.custom_attributes = other.custom_attributes
```
andre-senna: Is this a deep copy or a shallow one?

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
```diff
@@ -0,0 +1,118 @@
...
+            self.custom_attributes = custom_attributes
+            self.validate()
+
+    def __eq__(self, other: 'Atom') -> bool:
+        if not isinstance(other, Atom):
+            return False
+        return (
+            self.type == other.type
+            and self.is_toplevel == other.is_toplevel
+            and self.custom_attributes.P == other.custom_attributes.P
```
andre-senna: I don't understand what's happening here. What does this comparison with `.P` mean?

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
```diff
@@ -0,0 +1,118 @@
...
+        self,
+        type: str | None = None,
+        is_toplevel: bool = False,
+        custom_attributes: Properties = Properties(),
+        other=None
+    ) -> None:
+        if other:
+            self.type = other.type
+            self.is_toplevel = other.is_toplevel
+            self.custom_attributes = other.custom_attributes
```
andre-senna: Is this a deep copy or a shallow one?

3rd_party_slots/python_client/hyperon_das/commons/properties.py
```diff
@@ -19,10 +20,70 @@ def tokenize(self) -> list[str]:
             elif isinstance(v, bool):
                 vec.extend(["bool", str(v).lower()])
             elif isinstance(v, int):
-                vec.extend(["unsigned_int", str(v)])
+                if 0 <= v <= UINT32_MAX:
```
andre-senna: This doesn't work but I can't see a reasonable way to make it right. I believe we want to simplify our C++ class to use only unsigned int or long. For now, just add a comment here with a TODO.

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
marcocapozzoli: The `Properties` class has an attribute `P` which is a dictionary where properties are stored. The reason I compare `P` is that if I compare `self.custom_attributes == other.custom_attributes`, the two objects will have the same properties but will be different because they are different objects.
What I can do is implement __eq__ in the Properties class and remove this P.

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
marcocapozzoli: Shallow copy.
`self.custom_attributes = other.custom_attributes` copies only the reference to the same Properties object

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
marcocapozzoli: Shallow copy.
`self.custom_attributes = other.custom_attributes` copies only the reference to the same Properties object

andre-senna: So it's wrong. Here and in other parts of the code in Atom, Node and Link. Atom/Node/Link copy is supposed to be a deep copy. You need to be VERY careful with this. It's difficult to catch a bug here if you leave some behind.

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
andre-senna: I'm not sure I understood it. Perhaps my previous comment about shallow/deep copy may change the way we see this.

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
andre-senna: See similar comment above

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
marcocapozzoli: OK. I switched to a deep copy.

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
marcocapozzoli: I changed it.

3rd_party_slots/python_client/hyperon_das/commons/atoms/atom.py
marcocapozzoli: Not necessarily. Because I can have 2 nodes that have the same attributes, but if I compare them and don't have the `__eq__` in the `Properties` class, it's wrong because `custom_attributes` are different objects.

```
attr1 = Properties()
attr1['is_literal'] = True

attr2 = Properties()
attr2['is_literal'] = True

node1 = Node(type="Symbol", name='"A"', custom_attributes=attr1)
node2 = Node(type="Symbol", name='"A"', custom_attributes=attr2)`

assert node1 != node2
```

3rd_party_slots/python_client/hyperon_das/commons/properties.py
```diff
@@ -1,3 +1,4 @@
+UINT32_MAX = 2**32 - 1
 
 
 class Properties:
```
arturgontijo: This class must be a `dict`, eg:
```
class Properties(dict[str, str | int | float | bool]):
```
Or even better (`from collections import UserDict`):
```
class Properties(UserDict[str, str | int | float | bool]):
```

arturgontijo (changes requested):

arturgontijo (approved):

Issue #851, opened 2025-11-03
[MorkDB] Improve `query_for_targets()`


arturgontijo:
We currently get Atom's targets from MongoDB.
We should benchmark this approach against trying to fetch that info directly from MORK server.

arturgontijo:
Solved #852

PR #852, opened 2025-11-03
Add `MorkDB::query_for_targets()`

arturgontijo:
This PR implements the `query_for_targets()` for the MorkDB class.

We will just fetch the targets from the MongoDB for now. A next step would be improving this design to use Mork only (issue https://github.com/singnet/das/issues/851).

### DIFF
--- a/src/atomdb/morkdb/MorkDB.cc
+++ b/src/atomdb/morkdb/MorkDB.cc
@@ -184,15 +184,7 @@ shared_ptr<atomdb_api_types::HandleList> MorkDB::query_for_targets(const string&
     if (document == nullptr || !document->contains(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS])) {
         return nullptr;
     }
-
-    vector<string> targets;
-
-    unsigned int arity = document->get_size(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS]);
-    for (unsigned int i = 0; i < arity; i++) {
-        targets.push_back(string(document->get(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS], i)));
-    }
-
-    return make_shared<atomdb_api_types::HandleListMork>(targets);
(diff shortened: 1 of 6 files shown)


src/atomdb/morkdb/MorkDB.cc
```diff
@@ -179,6 +179,22 @@ shared_ptr<atomdb_api_types::HandleSet> MorkDB::query_for_pattern(const LinkSche
...
+shared_ptr<atomdb_api_types::HandleList> MorkDB::query_for_targets(const string& handle) {
+    auto document = this->get_atom_document(handle);
+    if (document == nullptr || !document->contains(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS])) {
+        return nullptr;
+    }
+
+    vector<string> targets;
+
+    unsigned int arity = document->get_size(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS]);
+    for (unsigned int i = 0; i < arity; i++) {
```
andre-senna: I dislike this way of doing it because it involves unnecessary copies being made.

Perhaps the best way is to have a constructor in `HandleListMork` that expects a reference to `document->get_size(MONGODB_FIELD_NAME[MONGODB_FIELD::TARGETS])`. If this is not possible or if you think it isn't proper, you should, at least, change the signature of `HandleListMork(vector<string> targets);` to `HandleListMork(vector<string>& targets);` to avoid copying the local `targets` vector.

andre-senna (changes requested):

src/atomdb/morkdb/MorkDB.cc
arturgontijo: We are building the `HandleListMork` from a `AtomDocument` ref now:
```
HandleListMork::HandleListMork(const shared_ptr<AtomDocument>& document) : HandleList() { ... }
```

andre-senna (approved):

Issue #821, opened 2025-10-24
Refactory `main` folder


andre-senna:
* Create 2 separate folders for architectural elements and test clients
* In the former, create a basic class with common argument parsing, signal;ing management and generic processor startup
* create separate main files for each agent reusing the basic class and adding specific parameters

PR #976, opened 2025-12-02
[FIX] Add methods to support Context Broker client

eddiebrissow:


andre-senna:
Related to https://github.com/singnet/das/issues/821

andre-senna (approved):

PR #981, opened 2025-12-03
[#960] Add delete atoms to AtomDB Broker

### LINKED ISSUE #960: Update and extend AtomDBProxy client API
Update current add_atoms() command to be compliant with recent changes in AtomDB and add a new delete_atoms() method as in the AtomDB API:

```
    virtual vector<string> add_atoms(const vector<atoms::Atom*>& atoms,
                                     bool throw_if_exists = false,
                                     bool is_transactional = false) = 0;
    
    virtual uint delete_atoms(const vector<string>& handles, bool delete_link_targets = false) = 0;
```

eddiebrissow:
closes #960

### DIFF
--- a/src/agents/atomdb_broker/AtomDBProxy.cc
+++ b/src/agents/atomdb_broker/AtomDBProxy.cc
@@ -88,7 +88,7 @@ vector<string> AtomDBProxy::add_atoms(const vector<Atom*>& atoms, bool use_strea
     // should be re-enabled by default.
 
     if (use_streaming) {
-        set_stream(AtomDBProxy::START_STREAM, stream_info);
+        this->to_remote_peer(AtomDBProxy::START_STREAM, stream_info);
     }
 
     for (Atom* atom : atoms) {
@@ -105,15 +105,12 @@ vector<string> AtomDBProxy::add_atoms(const vector<Atom*>& atoms, bool use_strea
         args.insert(args.begin(), atom_type);
         handles.push_back(atom->handle());
         if ((args.size() > AtomDBProxy::COMMAND_SIZE_LIMIT) || (atom == atoms.back())) {
-            Utils::retry_function([&]() { this->to_remote_peer(AtomDBProxy::ADD_ATOMS, args); },
-                                  5,
-                                  2000,
-                                  "AtomDBProxy::add_atoms");
+            this->to_remote_peer(AtomDBProxy::ADD_ATOMS, args);
             args.clear();
         }
     }
     if (use_streaming) {
-        set_stream(AtomDBProxy::END_STREAM, {});
+        this->to_remote_peer(AtomDBProxy::END_STREAM, {});
     }
 
     return handles;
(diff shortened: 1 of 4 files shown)


src/agents/atomdb_broker/AtomDBProxy.cc
```diff
@@ -118,6 +119,24 @@ vector<string> AtomDBProxy::add_atoms(const vector<Atom*>& atoms, bool use_strea
...
+    vector<string> args;
+    // split handles into chunks to avoid exceeding command size limits
+    size_t chunk_size = AtomDBProxy::COMMAND_SIZE_LIMIT;
+    uint total_deleted = 0;
+    for (size_t i = 0; i < handles.size(); i += chunk_size) {
+        args.clear();
+        size_t end = std::min(i + chunk_size, handles.size());
+        args.insert(args.end(), handles.begin() + i, handles.begin() + end);
+        args.push_back(delete_link_targets ? "1" : "0");
+        Utils::retry_function([&]() { this->to_remote_peer(AtomDBProxy::DELETE_ATOMS, args); },
```
andre-senna: This should be moved up to the proxy superclass so all proxies could share the same behavior.

src/agents/atomdb_broker/AtomDBProxy.cc
```diff
@@ -161,6 +183,20 @@ void AtomDBProxy::handle_add_atoms(const vector<string>& tokens) {
     }
 }
 
+void AtomDBProxy::handle_delete_atoms(const vector<string>& args) {
```
andre-senna: The name `handle_delete_atoms()` would be OK if we didn't use the term `handle` so intensively all along the code meaning something totally different. I suggest changing it. Perhaps using the terms "callback" or "piggyback". BTW I understand you just kept the naming convention which was already in the class. So perhaps we should change `handle_add_atoms()` as well.

marcocapozzoli (approved):

src/agents/atomdb_broker/AtomDBProxy.cc
```diff
@@ -104,20 +105,31 @@ vector<string> AtomDBProxy::add_atoms(const vector<Atom*>& atoms, bool use_strea
...
+    size_t chunk_size = AtomDBProxy::COMMAND_SIZE_LIMIT;
+    uint total_deleted = 0;
+    for (size_t i = 0; i < handles.size(); i += chunk_size) {
+        args.clear();
+        size_t end = std::min(i + chunk_size, handles.size());
+        args.insert(args.end(), handles.begin() + i, handles.begin() + end);
+        args.push_back(delete_link_targets ? "1" : "0");
+        this->to_remote_peer(AtomDBProxy::DELETE_ATOMS, args);
+    }
+    return total_deleted;
```
andre-senna: total_deleted is not being updated

andre-senna: Anyway, I'd just remove this variable and make this function return void.

andre-senna (approved):

PR #1007, opened 2026-01-16
[FIX] Fix Inference test code parameters

eddiebrissow:


src/tests/main/implication_query_evolution_main.cc
```diff
@@ -944,6 +944,8 @@ static void run(const string& predicate_source,
         AttentionBrokerClient::save_context(context, CONTEXT_FILE_NAME);
         exit(0);
     } else {
+        update_attention_allocation(predicate_source, predicate_target, context);
+        Utils::sleep(5000);
```
andre-senna: Why do we need this sleep?

andre-senna (approved):

PR #1016, opened 2026-01-26
[1/2] Implement `InMemoryDB`

arturgontijo:
This PR implements `InMemoryDB`, a new `AtomDB` class that stores all data in memory.

Note: There are still some methods to implement.


src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+// Helper class to wrap Atom in HandleTrie
+class AtomTrieValue : public HandleTrie::TrieValue {
+   public:
+    AtomTrieValue(Atom* atom) : atom_(atom) {}
+    ~AtomTrieValue() override { delete atom_; }
+    void merge(HandleTrie::TrieValue* other) override {
+        // For now, just replace (could be enhanced later)
+        delete atom_;
+        atom_ = dynamic_cast<AtomTrieValue*>(other)->atom_;
+        dynamic_cast<AtomTrieValue*>(other)->atom_ = nullptr;  // Prevent double delete
```
andre-senna: NULL is supposed to be used here instead of nullptr because atom_ is a raw pointer

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+
+// Helper class to store sets of atom handles in HandleTrie for pattern indexing
+class PatternTrieValue : public HandleTrie::TrieValue {
+   public:
+    PatternTrieValue() {}
+    ~PatternTrieValue() override {}
+    void merge(HandleTrie::TrieValue* other) override {
+        // Merge sets when the same pattern handle is inserted multiple times
+        PatternTrieValue* other_value = dynamic_cast<PatternTrieValue*>(other);
+        if (other_value != nullptr) {
```
andre-senna: nullptr --> NULL

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+    HandleSetInMemory* handle_set;
+};
+
+struct ReIndexData {
+    InMemoryDB* db;
+};
+
+bool re_index_visitor(HandleTrie::TrieNode* node, void* data) {
+    ReIndexData* index_data = static_cast<ReIndexData*>(data);
+    if (node->value != nullptr) {
```
andre-senna: I suggest making a revision of all nullptrs to check if they should be NULL instead.

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+        dynamic_cast<PatternTrieValue*>(this->incoming_sets_trie_->lookup(handle));
+    if (incoming_set_trie_value != nullptr) {
+        for (const auto& link_handle : incoming_set_trie_value->get_handles()) {
+            handle_set->add_handle(link_handle);
+        }
+    }
+    return handle_set;
+}
+
+// Stub implementations for AtomDocument methods (to be implemented later)
```
andre-senna, line 199: I'm OK with the stubs but they should raise an exception instead.

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+    return {};
+}
+
+vector<shared_ptr<AtomDocument>> InMemoryDB::get_matching_atoms(bool is_toplevel, Atom& key) {
+    // TODO: Implement in second phase
+    return {};
+}
+
+bool InMemoryDB::atom_exists(const string& handle) {
+    lock_guard<mutex> lock(trie_mutex_);
```
andre-senna: HandleTrie is already thread safe. Why do you see the need of having another mutex layer?

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+    for (const auto& handle : handles) {
+        if (this->delete_link(handle, delete_link_targets)) {
+            deleted_count++;
+        }
+    }
+    return deleted_count;
+}
+
+void InMemoryDB::re_index_patterns(bool flush_patterns) {
+    lock_guard<mutex> trie_lock(this->trie_mutex_);
```
andre-senna: This is an anti-pattern which can lead to a deadlock. Regard my previous comment regarding the necessity of adding an extra layer of mutex to manipulate HandleTrie. But anyway, even if you really want the extra layer, locking these two mutex this way is not safe. Suppose the following scenario:

1. Thread 1 call re_index_patterns() and acquire mutex1
2. Thread 2 acquire mutex2 by calling another method which locks only in mutex2
3. Thread 1 try to acquire mutex2 and blocks
4. Thread 2 try to acquire mutex1 by calling re_index_patterns() or any other method which locks in mutex1. It blocks.

Thread 1 and thread 2 are now in deadlock.

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -0,0 +1,824 @@
...
+    LinkSchema link_schema(tokens_vector);
+
+    this->pattern_index_schema_map[this->pattern_index_schema_next_priority] =
+        make_tuple(move(tokens_vector), index_entries);
+    this->pattern_index_schema_next_priority++;
+}
+
+vector<string> InMemoryDB::match_pattern_index_schema(const Link* link) {
+    vector<string> pattern_handles;
+    auto local_map = this->pattern_index_schema_map;
```
andre-senna: This operation performs a deep copy. is this the intended behavior?

arturgontijo: Lmk if the current approach is better.

andre-senna (approved):

PR #1025, opened 2026-02-04
[#1013]  Implement the Database Adapter

### LINKED ISSUE #1013: Implement the Database Adapter in C++

marcocapozzoli:
This PR adds the initial implementation of the Database Adapter. It introduces the abstract interfaces that define the adapter contract and provides a concrete Postgres-backed adapter implementation.

Resolves #1013


src/db_adapter/DBWrapper.h
```diff
@@ -0,0 +1,98 @@
...
+namespace db_adapter {
+
+/**
+ * @class DatabaseWrapper
+ * @brief Generic interface for a database connection wrapper.
+ *
+ * @tparam ConnT The underlying connection object type (e.g., pqxx::connection).
+ */
+template <typename ConnT>
+class DatabaseWrapper {
```
andre-senna, line 22: @eddiebrissow  @marcocapozzoli
Please make this class implement commons/processor/Processor.h. Basically it means you need to implement
    virtual void setup();
    virtual void start();
    virtual void stop();


setup() Is supposed to make proper initialization without starting any threads, connections etc. It's OK if it's empty.
start() is supposed to start any resources like threads, connections, file openings etc
stop() is supposed to gracefully stop the use of any resource. Stopping/synching threads, closing connections, closing open files etc.

The idea is that relevant resources like threads, connections etc become Processors as well. You can connect them (by calling add_subprocessor()) so when you setup/start/stop Processors they are setup/started/stopped in the proper order to avoid crashing.
We already have DedicatedThread.h and ThreadPool.h in the Processor package. The former is for when you need to spawn a single thread. The later is for when you want a thread pool of workers.
You may need to add a PostgresDBConnection (which inherits from Processor as well) in your package and probably you'll pass this class as the template argument instead of connection type.

src/db_adapter/DBWrapper.h
```diff
@@ -0,0 +1,98 @@
...
+   public:
+    explicit DatabaseWrapper(shared_ptr<Mapper> mapper, MAPPER_TYPE mapper_type)
+        : mapper(move(mapper)), mapper_type(mapper_type) {}
+    virtual ~DatabaseWrapper() = default;
+
+    /**
+     * @brief Closes the connection.
+     */
+    virtual void disconnect() = 0;
```
andre-senna, line 31: This method will disappear when you implement Processor

andre-senna (changes requested):

marcocapozzoli:
Another PR will replace this one.

PR #1034, opened 2026-02-09
[#1013] Implement the Database Adapter 2

marcocapozzoli:
This pull request adds the ContextLoader to validate the context file for mapping.

Resolves https://github.com/singnet/das/issues/1013


src/db_adapter/ContextLoader.h
```diff
@@ -0,0 +1,120 @@
...
+#include "Utils.h"
+
+using namespace std;
+using namespace commons;
+using namespace db_adapter;
+using json = nlohmann::json;
+
+namespace fs = std::filesystem;
+
+bool load_context_file(const string& file_path, vector<TableMapping>& out) {
```
andre-senna:
```suggestion
static bool load_context_file(const string& file_path, vector<TableMapping>& out) {
```

src/db_adapter/ContextLoader.h
```diff
@@ -0,0 +1,120 @@
...
+    out.clear();
```
andre-senna: Why? I don't see the point of doing this inside this function. If the caller wants a cleanup made upfront, it would make it itself before calling this method. Enforcing a cleanup here prevent the caller from calling the function twice for different input files in order to add more mappings to a previously computed vector of mappings. You may not do this right now but I don't see any reason to restrict the function this way.

src/db_adapter/ContextLoader.h
```diff
@@ -0,0 +1,120 @@
...
+
+    bool has_error = false;
+
+    if (!fs::exists(file_path)) {
+        LOG_ERROR("Context file " + file_path + " does not exist");
+        has_error = true;
```
andre-senna: Shouldn't the code return from here? What's the point of going on if the file don't exist?

src/db_adapter/ContextLoader.h
```diff
@@ -0,0 +1,120 @@
...
+        } else if (!table["table_name"].is_string()) {
+            LOG_ERROR(msg_base + ".table_name must be a string in a table entry");
+            has_error = true;
+        } else {
+            string tn = table["table_name"].get<string>();
+            size_t count_dot = 0;
+            for (char c : tn)
+                if (c == '.') ++count_dot;
+            if (count_dot != 1) {
+                LOG_ERROR(msg_base + ".table_name must be in format 'schema.table' (single dot)");
```
andre-senna:
```suggestion
                LOG_ERROR(msg_base + ".table_name must be in format 'schema.table'");
```

src/db_adapter/ContextLoader.h
```diff
@@ -0,0 +1,120 @@
...
+                has_error = true;
+            } else {
+                size_t pos = tn.find('.');
+                if (pos == 0 || pos + 1 >= tn.size()) {
+                    LOG_ERROR(msg_base + ".table_name parts must not be empty in a table entry");
```
andre-senna:
```suggestion
                    LOG_ERROR(msg_base + ".table_name must be in format 'schema.table'");
```

src/db_adapter/ContextLoader.h
```diff
@@ -0,0 +1,120 @@
...
+                    }
+                }
+            }
+        }
+        
+        out.push_back(tm);
+    }
+
+    if (has_error) {
+        LOG_ERROR("Context file validation failed with errors. Please fix the issues and try again.");
```
andre-senna: I understand you allowed the JSON processing continue after an error to allow the caller to see all syntax error messages at once instead of going one by one. If this is the case you aren't supposed to give any outputs when errors occur. So you need to change the logic of the way you output the mappings.

I suggest creating a local vector and here in this `if`, if no errors occurred you feed the contents of this internal vector to the output (without cleaning it up before). If errors occur, nothing is added to the output.

src/db_adapter/ContextLoader.h
andre-senna: You don't put actual code in `.h` files. Usually, the only exception is when you are declaring a templated class (which is not the case here).

To achieve what you intended here, I suggest creating a class ContextLoader (with a `.h` and a `.cc` files) and declaring load_context_file() as a static method (i.e. a class method).

andre-senna (changes requested):

src/db_adapter/ContextLoader.h
marcocapozzoli: > Enforcing a cleanup here prevent the caller from calling the function twice for different input files

Yes, you're right, I hadn't thought of that.

src/db_adapter/DataTypes.h
```diff
@@ -91,4 +91,10 @@ struct Table {
  */
```
andre-senna, line 91: These types are not attached to "db_adapter" but to SQL stuff. So these types should be defined wherever you define SQL stuff (probably PostgresWrapper.h)

marcocapozzoli:
Another PR will replace this one.

PR #1035, opened 2026-02-09
[#1013] Implement the Database Adapter

### LINKED ISSUE #1013: Implement the Database Adapter in C++

marcocapozzoli:
This PR adds Processors to the database connection. i.e, implement commons/processor/Processor.h

Resolves https://github.com/singnet/das/issues/1013


src/db_adapter/DBWrapper.h
```diff
@@ -3,42 +3,59 @@
 #include <memory>
```
andre-senna, line 3: Please name the file with the name of the main class defined in it: DatabaseWrapper

andre-senna (changes requested):
The code inside DBWrapper.{h,cc} is a mess. Although you can define multiple classes in the same files in C++ (and although this is a good choice sometimes) most of the time you should keep one pair of `.h` and `.cc` files per class. I suggest you to do it this way here, splitting the classes you create din this file because I believe you are over-templatizing the code.

So please break down this code in multiple files before submitting it to review.

src/db_adapter/DatabaseWrapper.cc
```diff
@@ -0,0 +1,24 @@
+#include "DatabaseWrapper.h"
+
+namespace {
```
andre-senna: What is this statement doing?

src/db_adapter/DatabaseConnection.cc
```diff
@@ -0,0 +1,47 @@
...
+
+DatabaseConnection::DatabaseConnection(const string& id, const string& host, int port) : Processor(id) {
+    this->host = host;
+    this->port = port;
+    this->connected = false;
+}
+
+DatabaseConnection::~DatabaseConnection() {}
+
+void DatabaseConnection::setup() {
```
andre-senna: If you have nothing to do in setup(), just remove it. You are replicating superclass implementation here.

src/db_adapter/postgres/PostgresWrapper.cc
```diff
@@ -13,46 +13,70 @@
...
+PostgresDatabaseConnection::PostgresDatabaseConnection(const string& id,
+                                                       const string& host,
+                                                       int port,
+                                                       const string& database,
+                                                       const string& user,
+                                                       const string& password)
+    : DatabaseConnection(id, host, port), database(database), user(user), password(password) {}
+
+PostgresDatabaseConnection::~PostgresDatabaseConnection() {
+    if (this->is_running()) {
```
andre-senna: This is replicating superclass behavior. Just let the destructor empty.

src/db_adapter/Pipeline.h
```diff
@@ -0,0 +1,66 @@
...
+                       MAPPER_TYPE mapper_type = MAPPER_TYPE::SQL2ATOMS,
+                       shared_ptr<SharedQueue> output_queue = nullptr);
+    ~DatabaseMappingJob() = default;
+
+    void add_task_query(const string& virtual_name, const string& query);
+    void add_task_table(TableMapping table_mapping);
+
+    bool thread_one_step() override;
+
+    bool is_finished() const;
```
andre-senna, line 35: You should rely on Processor's. is_finished(). You can't re-implement it.

src/db_adapter/Pipeline.h
```diff
@@ -0,0 +1,66 @@
...
+        enum Type { TABLE, QUERY } type;
+        TableMapping table_mapping;
+        string virtual_name;
+        string query;
+    };
+    unique_ptr<PostgresDatabaseConnection> db_conn;
+    unique_ptr<PostgresWrapper> wrapper;
+    vector<MappingTask> tasks;
+    size_t current_task = 0;
+    bool finished = false;
```
andre-senna, line 48: Don't replicate superclass behavior. Remove this flag and rely on superclass is_finished()

src/db_adapter/Pipeline.h
```diff
@@ -0,0 +1,66 @@
...
+    bool initialized = false;
```
andre-senna, line 49: Same comment.

src/db_adapter/Pipeline.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
+class AtomPersistenceJob : public ThreadMethod {
+   public:
+    AtomPersistenceJob(shared_ptr<SharedQueue> input_queue);
+    ~AtomPersistenceJob() = default;
+
+    bool thread_one_step() override;
+    bool is_finished() const;
```
andre-senna, line 58: Again

src/db_adapter/Pipeline.cc
```diff
@@ -0,0 +1,117 @@
...
+        Utils::error("Error in Worker: " + string(e.what()));
+        for (auto& atom : atoms) {
+            delete atom;
+        }
+    }
+    this->finished = false;
+    return false;
+}
+
+bool AtomPersistenceJob::is_finished() const { return this->finished; }
```
andre-senna, line 145: Remove this method

andre-senna (changes requested):

src/db_adapter/DatabaseWrapper.cc
marcocapozzoli: This is an anonymous namespace.
`create_mapper` can only be used within this .cc file.

src/db_adapter/Pipeline.h
marcocapozzoli: @andre-senna 
I'm not reimplementing the `Processor` method because this class is a `ThreadMethod`. The processor manages the resource, which in this case, the `DedicatedThread` manages the thread. Therefore, `is_finished` signals that the thread has stopped. But what I need is a mechanism that tells me when the `Job` has finished all its tasks. It just so happens that I named it with the same name, `is_finished`. In short, I think it shouldn't be removed because I need to know when the `Job` finishes.

src/db_adapter/Pipeline.h
marcocapozzoli: https://github.com/singnet/das/pull/1035#discussion_r2863963119

src/db_adapter/Pipeline.h
marcocapozzoli: https://github.com/singnet/das/pull/1035#discussion_r2863963119

src/db_adapter/Pipeline.h
marcocapozzoli: https://github.com/singnet/das/pull/1035#discussion_r2863963119

src/db_adapter/Pipeline.cc
marcocapozzoli: https://github.com/singnet/das/pull/1035#discussion_r2863963119

src/commons/HandleTrie.cc
```diff
@@ -26,7 +27,10 @@ HandleTrie::TrieNode::~TrieNode() {
         delete children[i];
     }
     delete[] children;
-    delete value;
+    if (value != NULL) {
```
ccgsnet, line 30: Let's add a `TODO` here so we can remove this check once we improve `HandleTrie::insert()` to not accept `NULL` as `value`

src/commons/HandleTrie.cc
```diff
@@ -265,3 +269,8 @@ void HandleTrie::traverse(bool keep_root_locked,
         root->trie_node_mutex.unlock();
     }
 }
+
+bool HandleTrie::exists(const string& key) {
+    TrieNode* node = lookup_node(key);
+    return node != NULL ? true : false;
```
ccgsnet, line 279: Suggestion:
```
return lookup_node(key) != NULL;
```

src/commons/HandleTrie.cc
```diff
@@ -3,6 +3,7 @@
 #include <iostream>
 #include <stack>
 
+#include "Logger.h"
```
ccgsnet: Unused

src/db_adapter/ContextLoader.cc
```diff
@@ -0,0 +1,151 @@
...
+    for (size_t i = 0; i < tables.size(); ++i) {
+        string msg_base = "table[" + to_string(i) + "]";
+
+        const json& table = tables[i];
+
+        TableMapping tm;
+
+        if (!table.contains("table_name")) {
+            LOG_ERROR(msg_base + " missing required key: 'table_name'");
+            has_error = true;
```
ccgsnet, line 40: Don't we want to break the `for` loop once validation has failed?
Why we would keep iterating when a single validation failure is not allowed?

src/db_adapter/DataMapper.cc
```diff
@@ -0,0 +1,414 @@
...
+
+        return "\"" + escaped + "\"";
+    }
+    return "";
+}
+
+bool BaseSQL2Mapper::insert_handle_if_missing(const string& handle) {
+    auto exists = this->handle_trie.exists(handle);
+    if (exists) return false;
+    this->handle_trie.insert(handle, NULL);
```
ccgsnet: I think we don't want to allow `NULL` here...
We should use the same approach that I used at the `RemoteAtomDBPeer.h` (the `EmptyTrieValue`).
Maybe we should move the `EmptyTrieValue` class to `commons` and use it here

src/db_adapter/DatabaseConnection.h
```diff
@@ -0,0 +1,35 @@
...
+    virtual void disconnect() = 0;
+
+    bool is_connected() const;
+
+   protected:
+    string host;
+    int port;
+
+   private:
+    bool connected;
```
ccgsnet, line 31: Doesn't this need to me atomic?
```
std::atomic<bool> connected
```

src/db_adapter/postgres/PostgresWrapper.h
```diff
@@ -0,0 +1,94 @@
...
+
+    Table get_table(const string& name) override;
+    vector<Table> list_tables() override;
+    void map_table(const Table& table,
+                   const vector<string>& clauses,
+                   const vector<string>& skip_columns = {},
+                   bool second_level = false) override;
+    void map_sql_query(const string& virtual_name, const string& raw_query) override;
+
+    mutex api_mutex;
```
ccgsnet: Does this need to be `public`?

src/db_adapter/postgres/PostgresWrapper.cc
```diff
@@ -0,0 +1,495 @@
...
+    : SQLWrapper(db_conn, mapper_type), db_conn(db_conn), output_queue(output_queue) {}
+
+PostgresWrapper::~PostgresWrapper() {}
+
+Table PostgresWrapper::get_table(const string& name) {
+    auto tables = this->list_tables();
+    for (const auto& table : tables) {
+        if (table.name == name) return table;
+    }
+    Utils::error("Table '" + name + "' not found in the database.");
```
ccgsnet, line 119: I think it would be better to not throw here, just log the error.

src/tests/cpp/db_adapter_test.cc
```diff
@@ -0,0 +1,877 @@
...
+
+            EXPECT_TRUE(conn->is_connected());
+
+            conn->execute_query("SELECT 1");
+
+            count_threads++;
+
+            conn->stop();
+
+            EXPECT_FALSE(conn->is_connected());
```
ccgsnet, line 240: Let's add a `delete conn` here and in the `catch` so we can release the connection.

src/tests/cpp/db_adapter_test.cc
```diff
@@ -0,0 +1,877 @@
...
+        if (table.name == FEATURE_TABLE) found_feature = true;
+    }
+
+    EXPECT_TRUE(found_organism);
+    EXPECT_TRUE(found_cvterm);
+    EXPECT_TRUE(found_feature);
+
+    vector<Table> tables_cached = wrapper->list_tables();
+
+    ASSERT_EQ(tables_cached.size(), tables_cached.size());
```
ccgsnet:
```
ASSERT_EQ(tables_cached.size(), tables.size())
```

src/tests/main/database_adapter_main.cc
```diff
@@ -0,0 +1,158 @@
...
+#define LOG_LEVEL INFO_LEVEL
+#include "Logger.h"
+
+using namespace std;
+using namespace commons;
+using namespace atomdb;
+using namespace db_adapter;
+using namespace processor;
+
+void ctrl_c_handler(int) {
```
ccgsnet, line 22: As we are dealing with DBs, it would be better to be sure all operations were done before `exit(0)` so we don't corrupt any data.

src/db_adapter/ContextLoader.cc
marcocapozzoli: So that the user can check all the errors I may make at once, instead of fixing them one by one.

andre-senna (approved):

Issue #1009, opened 2026-01-19
Implement an evolutionary algorithm to prove hidden relationship between a given predicate and a given concept


andre-senna:
We want to prove the existence of an expression $(Evaluation \\> P_{target} \\> C_{target})$.

$(Evaluation \\> P_{target} \\> C_{target})$ is not present in the knowledge base but it's in the "test" knowledge base, i.e., we know a priori it's true.

Since $(Evaluation \\> P_{target} \\> C_{target})$ is not present in the knowledge base, we need to find a combination of expressions which implies in $(Evaluation \\> P_{target} \\> C_{target})$. We'l do so by making a query to the QueryEvolution agent. 

# High level algorithm

We'll apply a query evolution request in an iterative algorithm which also includes complementary link creation. These are the steps of the algorithm:

1. Initialize ContextBroker
2. Pre-processing
3. Run a round of complementary links creation
4. Run query evolution
5. Repeat steps 3. and 4. until a stop criterium is reached

## ContextBroker initialization

ContextBroker must be initialized to make Evaluation importance depends on Concept and Predicate importance's.

```
LINK_TEMPLATE 3
    Evaluation
    VARIABLE P1
    VARIABLE C1

{(toplevel_link, P1), (toplevel_link, C1)}
```

There's no need to spread initial activation here, so the activation spreading parameter can be set empty.

## Pre-processing

In the pre-processing step we must accomplish two tasks: (1) initialization of STI and (2) creation of basic complementary links.

(1) must bump STI of all concepts and predicates linked to the PTARGET and CTARGET via some Evaluation expression:

```
OR 2
    LINK_TEMPLATE 3
        Evaluation
        ATOM PTARGET
        VARIABLE C1
    LINK_TEMPLATE 3
        Evaluation
        VARIABLE P1
        ATOM CTARGET
```

(2) Should re-use the link-creation code to compute "equivalence" between CTARGET and every other concept in the knowledge base. Then we must do the same for PTARGET computing "implication" from/to all the other predicates in the knowledge base. But, for "implication", we must consider equivalent concepts (with the proper strength being taken into account) as well as actual concepts linked to a given predicate.

## Link creation

We'll create 3 types of complementary links.

1. Equivalence
2. Implication
3. Evaluation

1. and 2. must follow the same implementation we have in our current code. As for 3., the idea is to select two evaluation links pointing to the same concept and create a new Evaluation with the conjunction of the two predicates pointing to the same concept. So the query would be like this:

```
AND 2
[...]

PR #1036, opened 2026-02-10
[#1009] Joint test client for inference and query evolution

andre-senna:
This is a new test client for inference and query evolution. The idea is to have a simplified inference algorithm which uses query evolution to evolve a chain of links in order to prove a premise.

This client is supposed to be used to solve both Rejuve-Flybase and Toy inference problem. To use it in the later, it's still required to fix some of the queries (e.g. for context creation).

Minor fixes e.g. in the Query Engine have also been added in order to make the client work. These are listed below.

* ContextBroker: added a new proxy parameter DISREGARD_IMPORTANCE_FLAG which is passed to the query engine in order to define if STI is supposed to be used or not in the query.
* QueryEvolutionProcessor: fixed the way correlation query elements are replaced by query answer elements. The previous code was made fine-tuned to the word toy problem and were not generic. The way it is in this PR, this code is (hopefully) general enough to be used in any problem.
* LinkSchema: added a new constructor parameter which is a flag to optionally disregard STI when fetching query answers
* LinkSchema: added a logic to determine if the LinkTemplate is a "flat" pattern (i.e. has no nested link templates). This is just a preparation to a code that will be made in a follow-up PR which will use this information in order to avoid fetching data from MongoDB when the pattern is flat.
* LinkSchema: added a INFO log message to show progress on processing of long queries.
* MettaParserActions, PatternMatcherQuery*: support for this new disregard_importance flag
* Multiple tests have also been updated because of this change in LinkTemplate constructor


src/agents/query_engine/PatternMatchingQueryProcessor.cc
```diff
@@ -276,6 +277,9 @@ shared_ptr<QueryElement> PatternMatchingQueryProcessor::setup_query_tree(
         } else {
             Utils::error("Invalid token in query: " + query_tokens[cursor]);
         }
+        if ((query_tokens[cursor] == LinkSchema::LINK_TEMPLATE) || (query_tokens[cursor] == OR) ||
+            (query_tokens[cursor] == AND)) {
```
eddiebrissow: Missing logic inside if

eddiebrissow (approved):

src/agents/query_engine/PatternMatchingQueryProcessor.cc
andre-senna: Dead code. Fixed.

ccgsnet (approved):
LGTM

Issue #1037, opened 2026-02-10
Implement a concrete AtomDB to support remote connection to multiples AtomDBs


andre-senna:
## **This issue description is a WIP. The text is not ready to be used yet.** 

This is also related to #915. 

The high level idea is to allow the user to setup a DAS which, instead of having a DB backend like redis+mongo or mork+mongo, is backed by a list of remote DASes, each one with its own DB backend.

In order to instantiate such a DAS, the user will need to select a new concrete AtomDB type (the one being defined here) which is a RemoteAtomDB. So this idea is to delegate to this RemoteAtomDB everything that's required in order to manipulate data from the various remote DASes.

So what we need to do here is to design how this RemoteAtomDB should be/work. Lets make a pseudo C++ code for it:

```

class RemoteAtomDB : public AtomDB {

    // Constructor expects a JSON file with everything required to configure the RemoteAtomDB.
    // Basically this means:
    //     * connection information for each of the remote DBs it's going to connect to.
    RemoteAtomDB(const string& json_file_path);

    // Remote AtomDBs this AtomDB is connected to. Instead of
    // using AtomDB objects directly here, we use RemoteAtomDBPeer objects
    // because we need to keep more state regarding each remote DB
    // other than the AtomDB itself. This RemotePeer is basically what we
    // have been calling "CachedAtomDB".
    //
    // Each RemoteAtomDBPeer should have a UID based in the remote AtomDB's connection info.
    map<string, RemoteAtomDBPeer> remote_db; 
};

class RemoteAtomDBPeer : public AtomDB {

    // Each remote peer has 3 AtomDB objects:
    //    * One to held the remote AtomDB itself
    //    * One in-memory cache
    //    * One regular DB-backed (e.g. redis+mongo) AtomDB to persist newly added atoms.

    // ---------------------------------------------
    // State
 
    // Cache with relevant Atoms. The idea is that all public AtomDB query API should be redirected to
    // this cache except for query_for_patterns() which should redirected to the AtomDB below.
    // In order to be able to answer for these queries, the cache must be properly fed up with relevant
    // atoms. 
    InMemoryDB cache;

    // Remote DB being connected to. This AtomDB is supposed to be used to READ-ONLY.
    // query_for_patterns() is delegated to this AtomDB.
    AtomDB atomdb;
    
    // Any new atom addition/deletion should be redirected to local_persistence as well as to the AtomDB above.
    AtomDB local_persistence;

[...]

PR #1044, opened 2026-02-19
Add `RemoteAtomDB` and `RemoteAtomDBPeer`

ccgsnet:
Here are some design decisions that we still need to discuss:

Reference: https://github.com/singnet/das/issues/1037

1 - `RedisMongoDB` and `MorkDB` rely on ENV VARs to connect to the external DBs services (eg. `DAS_REDIS_HOSTNAME`, `DAS_MONGODB_HOSTNAME`, `MORK_URL`, etc). We can pass that info in the `RemoteAtomDB` `config.json` file but we would still need to change the constructors of those classes.

2 - `available_ram()` – Returns raw KB (from `Utils::get_current_free_ram()`) instead of a fraction in (0,1). Maybe we could pass the amount of total RAM to be used by the `RemoteAtomDBPeer` in the `config.json`. That way we can do the math and return the 0 - 1 value.

3 - `auto_cleanup()` – To be implemented.

4 - `add_*` semantics – Doc says `redirect calls to cache`. Should we also writes to `local_persistence`? This matches the goal of persisting new atoms.

5 - `query_for_pattern()` – Current implementation merges `atomdb + local_persistence + cache`. Should we really redirect just to `atomdb` when schema not yet fetched?

6 - `query_for_targets() and query_for_incoming_set()` – Doc says we must use cache-only. Current implementation does full fallback (cache → local_persistence → atomdb).

7 - `get_matching_atoms(key, local_only=false)` – No public API for `local_only=true` case yet.

8 - Peer `uid`: we need a way to set (eg, via class variable or function param) the UID of the target peer.
```
remote_db.set_peer("peer_uid_1");
bool result = remote_db.link_exists(link_handle);
// run more operations against "peer_uid_1"
remote_db.set_peer("peer_uid_2");
// run operations against "peer_uid_2"
```
9 - Add/Remove new `RemoteAtomDBPeer`.


andre-senna:
- 1 - 

Using env vars for this was a bad design decision since the beginning but we never took the time to fix this. This point has been an annoyance and a source of errors since ever. My suggestion is that you hack RemoteAtomDB to create the internal AtomDBs using the config file as we discussed but without changing the constructors as you said. You should call setenv() instead before calling the constructor. If you want to be more cautious, you could (a) call getenv() before, saving any values already set in the user's environment, then (b) call setenv() to set the values passed in the config file, then (c) call the constructors, which will use the proper values in the env vars and, finally, (d) call setenv() again, passing the values you saved previously.

andre-senna:
- 2 -

Passing absolute amount of RAM to be used in the config JSON is OK but I still prefer to pass a % RAM. Using a % the config file would be more robust i.e. preventing the user from keeping a very low value when "migrating" a test from a local workstation to a more powerful server or, worst, preventing DAS from depleting some server's RAM because the config file hasn't been changed when the user migrated the test from a more powerful to a more modest server.

My suggestion is to add a new static function to Utils to get the whole machine RAM.

andre-senna:
* 4 - 

No. Persistence shouldn't be automatic. Persistence of newly created atoms should be explicitly requested by caller. This is because most of the atoms created during the execution of an AI algorithm are supposed to be temporary only, just to allow the algorithm to reach a goal. After that goal is reached, part of the created atoms (the part that lead to the good answers) are kept and part (the part that didn't lead anywhere useful) are just discarded.

andre-senna:
- 5 -

> Current implementation merges atomdb + local_persistence + cache

This sounds incorrect. Can you explain in more details what you are doing?

 > Should we really redirect just to atomdb when schema not yet fetched?
 
 You asked "just to atomddb". I don't understand. Where else could you redirect it if it isn't fetched yet? If you actually meant "Should we really redirect to atomdb just when schema is not yet fetched?" the answer is YES.

andre-senna:
- 6 - 

I didn't find these in the description of #1037  . Where is this described?

andre-senna:
- 8 - 

Ugh! :-P

When these methods gets to searching in the atomdb (i.e. when they failed to find the atom in cache and persistence) they are supposed to search in all remote AtomDBs.

Perhaps we may want to add an extra set of methods (which will not go to the abstract AtomDB API) which expects an extra parameter of the UID of the AtomDB that must be searched but I prefer to add this IF we hit a concrete use case.

andre-senna:
- 9 - 

You didn't write any comments. Does it require discussion or it's just not implemented yet?

andre-senna:
Regarding 6. @ccgsnet  said:

> Current implementation does full fallback (cache → local_persistence → atomdb).

This is correct. This is how it's suppose to work indeed.

andre-senna:
Regarding 8.

Please disregard my previous comment. There's no need to consider multiple remote AtomDBs here. We need to focus on the `RemoteAtomDBPeer` design for now. The matter of multiple atom DBs should be taken care of when we work the design for `RemoteAtomDB` API

src/atomdb/remotedb/RemoteAtomDB.cc
```diff
@@ -0,0 +1,403 @@
...
+using namespace atoms;
+using namespace commons;
+
+using json = nlohmann::json;
+
+namespace {
+
+// Saves current env vars for keys in config["env"], then setenv's from config.
+// Restorer restores them in destructor (or when restore() is called).
+class EnvRestoreGuard {
```
andre-senna: This should be renamed to EnvironmentRestoreGuard and moved to the commons package.

src/atomdb/remotedb/RemoteAtomDB.cc
```diff
@@ -0,0 +1,403 @@
...
+        Utils::error("RemoteAtomDB: Invalid JSON in config: " + string(e.what()));
+    }
+
+    auto doc = doc_opt->at("atomdbs");
+    for (auto&& peer_config : doc) {
+        string uid = peer_config.at("uid");
+
+        shared_ptr<AtomDB> remote = create_atomdb_from_config(peer_config);
+
+        shared_ptr<AtomDB> local_persistence;
```
andre-senna:
```suggestion
        shared_ptr<AtomDB> local_persistence = nullptr;
```

src/atomdb/remotedb/RemoteAtomDB.cc
```diff
@@ -0,0 +1,403 @@
...
+        if (peer_config.contains("local_persistence")) {
+            auto local_it = peer_config.at("local_persistence");
+            if (local_it.is_object()) {
+                local_persistence = create_atomdb_from_config(local_it);
+            }
+        }
+        if (!local_persistence) {
```
andre-senna:
```suggestion
        if (local_persistence == nullptr) {
```

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    return nullptr;
+}
+
+vector<shared_ptr<Atom>> RemoteAtomDBPeer::get_matching_atoms(bool is_toplevel, Atom& key) {
+    return get_matching_atoms_impl(is_toplevel, key, false);
+}
+
+vector<shared_ptr<Atom>> RemoteAtomDBPeer::get_matching_atoms_impl(bool is_toplevel,
```
andre-senna: This method is, actually, part of the public API. It's weird having the suffix "_impl" in its name. And it's not required. Do you see any reason to not reusing the name get_matching_atoms()?

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+
+        string handle(handle_cstr);
+        auto atom = get_atom(handle);
+        if (atom) {
+            cache_.add_atom(atom.get());
+        }
+    }
+}
+
+shared_ptr<HandleSet> RemoteAtomDBPeer::query_for_pattern(const LinkSchema& link_schema) {
```
andre-senna: The issue description for this method is wrong. My bad, I apologize. Actually, if the schema is already fetched, the answer may be either in cache or in persistence. So when schema is already fetched, both should be queried and results should be merged properly.

If schema is not fetched yet, only the atomdb should be queried. In this case, results should be added to cache, as you did.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    if (atomdb_) {
+        return atomdb_->query_for_targets(handle);
+    }
+
+    return nullptr;
+}
+
+shared_ptr<HandleSet> RemoteAtomDBPeer::query_for_incoming_set(const string& handle) {
```
andre-senna, line 211: The issue card don't mention this method. My mistake again. Anyway, retrieving the incoming set is not as trivial as the targets of a link. The same atom may have links pointing to them in the cache, in persistence and in the remote atomdb.

Therefore, instead of returning after searching each level, you are supposed to check ALL levels (cache, persistence and atomdb) and merge the results.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+        local_persistence_->re_index_patterns(flush_patterns);
+    }
+}
+
+void RemoteAtomDBPeer::fetch(const LinkSchema& link_schema) {
+    if (!schema_already_fetched(link_schema) && atomdb_) {
+        auto result = atomdb_->query_for_pattern(link_schema);
+        if (result) {
+            feed_cache_from_handle_set(result);
+            fetched_link_templates_.insert(link_schema.handle(), new EmptyTrieValue());
```
andre-senna: Is it really required to instantiate a new EmptyTrieValue object every time? I understand we could use the same object always.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+        }
+    }
+}
+
+void RemoteAtomDBPeer::release(const LinkSchema& link_schema) {
+    fetched_link_templates_.remove(link_schema.handle());
```
andre-senna: release() should make a query to cache and save the results to persistence after removing the results from cache. This is not 100% correct because we may remove atoms required/fetched by another schema but we shaw live with this limitation for now.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+}
+
+double RemoteAtomDBPeer::available_ram() { return static_cast<double>(Utils::get_current_free_ram()); }
```
andre-senna: This method is supposed to return a number in [0,1] which is basically Utils::get_current_free_ram()/Utils::get_total_ram()

get_total_ram() is not implemented but it should be an easy one.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -0,0 +1,459 @@
...
+// TODO: Implement cache cleanup properly
+void RemoteAtomDBPeer::auto_cleanup() {
+    auto available_ram_mb = available_ram() / 1024;
+    if (available_ram_mb < 128) {
+        LOG_INFO("RemoteAtomDBPeer " << uid_ << ": Low RAM detected (available=" << available_ram_mb
+                                     << " MB), cache cleanup triggered");
+    }
+}
+
+void RemoteAtomDBPeer::start_cleanup_thread() {
```
andre-senna, line 456: Is this being actually used? If no, let's remove it. If yes, we need to use commons/processor/DedicatedThread instead of using std:thread directly.

andre-senna (changes requested):

src/atomdb/remotedb/RemoteAtomDBPeer.cc
ccgsnet: `HandleTrie::remove()` calls `delete node->value`, so the second call to `remove()` would crash.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
ccgsnet: Will keep it (but using `DedicatedThread`) as we will have the proper `auto_cleanup()` implementation in a follow up PR.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
andre-senna: Perhaps we may add an optional flag defaulted to true to tell if value is supposed to be deleted or not.

andre-senna (approved):

andre-senna (approved):

Issue #212, opened 2025-01-24
Implement CHAIN operator in the Query Agent's query tree


andre-senna:
```
REP
    OR
        Eq PREV V
        Imp PREV V
    A
    B
    n
```

Translates to (n == 2):

```
OR
    OR
        Eq A B
        Imp A B
    AND
        OR
            Eq A V1
            Imp A V1
        OR
            Eq V1 B
            Imp V1 B
    AND
        OR
            Eq A V1
            Imp A V1
        OR
            Eq V1 V2
            Imp V1 V2
        OR
            Eq V2 B
            Imp V2 B
```

andre-senna:
REP renamed to CHAIN

PR #1049, opened 2026-03-04
[#212] Implementation of CHAIN operator in pattern matching query

andre-senna:
WIP towards #212 

Still missing the integration of the operator with the query parse. I.e. for now, it's not possible to use the query client or MeTTa to put CHAIN operators in query. This PR covers only the implementation of the QueryElement.

The idea of this operator is to allow the caller to search for a chain of links connecting a SOURCE to a TARGET.  This is another operator to be used in a query (like AND, OR etc). Whenever a CHAIN operator appears in a query tree (together with a SOURCE and a TARGET parameters) the operator will use its input to search for paths connecting SOURCE and TARGET.

Paths are passed through to the next query element in the query tree in QueryAnswers with N handles meaning a path of size N where source is the first handle and target is the last one.

Depending on an optional parameter, incomplete paths (connected only to either the source or the target) are also forwarded.

Paths are searched in a combinatorial fashion but STI is used to direct the search through the search space in a way that paths with the largest STI are searched and reported earlier.

In this version, the input links are supposed to be ternary. Something like `(Similarty $v1 $v2)`. Actually, the first target (`Similarituy` in this example) is totally disregarded.


src/agents/query_engine/query_element/Chain.cc
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// --------------------------------------------------------------------------------------------
+// Private stuff
+
+void Chain::initialize(const array<shared_ptr<QueryElement>, 1>& clauses) {
+    if (clauses.size() != 1) {
+        Utils::error("Invalid Chain operator with " + std::to_string(clauses.size()) + " clauses.");
+    }
+    this->id = "CHAIN(" + clauses[0]->id, +", " + this->source_handle + ", " + this->target_handle + ")";
```
eddiebrissow:
```suggestion
    this->id = "CHAIN(" + clauses[0]->id + ", " + this->source_handle + ", " + this->target_handle + ")";
```

src/agents/query_engine/query_element/Chain.cc
```diff
@@ -0,0 +1,415 @@
...
+                      << "New query answer: " << answer->to_string());
+            for (string handle : answer->handles) {
+                auto iterator = this->known_links.find(handle);
+                if (iterator == this->known_links.end()) {
+                    this->known_links.insert(iterator, handle);
+                    shared_ptr<Link> link =
+                        dynamic_pointer_cast<Link>(AtomDBSingleton::get_instance()->get_atom(handle));
+                    if (link == nullptr) {
+                        LOG_DEBUG("[CHAIN OPERATOR] "
+                                  << "NULL link");
```
eddiebrissow: Missing a `continue`or exception throw here.

eddiebrissow (approved):

ccgsnet (approved):

src/agents/query_engine/query_element/Chain.cc
andre-senna: Fixed

src/agents/query_engine/query_element/Chain.cc
andre-senna: Fixed

PR #1059, opened 2026-03-11
[DAS#1056] Setup script for tests-all (cache/no-cache)

levisingularity:
* Added a setup script, initializes DBs and knowledge bases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,12 +100,10 @@ test-clear:
 
 test-all-no-cache: setup-test-all
 	@$(MAKE) bazel 'test --show_progress --cache_test_results=no //tests/...'
-	@$(MAKE) test-clear
 
 test-all: 
 	@$(MAKE) setup-test-all
 	@$(MAKE) bazel 'test --show_progress //tests/...'
-	@$(MAKE) test-clear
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -89,12 +89,23 @@ clean:
...
+
+test-all-no-cache: setup-test-all
 	@$(MAKE) bazel 'test --show_progress --cache_test_results=no //tests/...'
+	@$(MAKE) test-clear
 
-test-all: build-image run-tests-db-loader
+test-all: 
+	@$(MAKE) setup-test-all
 	@$(MAKE) bazel 'test --show_progress //tests/...'
+	@$(MAKE) test-clear
```
andre-senna: test-all shouldn't call test-clear. Users is expected to call test-clear explicitly if they want to clear the containers.

andre-senna (approved):

Makefile
levisingularity: https://github.com/singnet/das/pull/1059/commits/e335f1530ede47b84bea4dfeec44157f90b4e394

Removed from test-all commands, now it's optional for the user.

PR #1070, opened 2026-04-13
[#1067] Database Adapter  multiple threads

### LINKED ISSUE #1067: Database adapter with support for multiple threads.

marcocapozzoli:
Resolves #1067


src/db_adapter/BoundedSharedQueue.h
```diff
@@ -0,0 +1,29 @@
...
+
+    void enqueue(void* item);
+    void* dequeue();
+    bool empty();
+    size_t size();
+
+   private:
+    queue<void*> queue_;
+    mutex mtx_;
+    condition_variable not_empty_;
```
andre-senna: I didn't see where this is used

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,306 @@
...
+        for (size_t i = 0; i < BATCH_SIZE; ++i) {
+            if (this->input_queue->empty()) break;
+            auto atom = (Atom*) this->input_queue->dequeue();
+            this->count++;
+            if (atom == nullptr) {
+                Utils::error("Dequeued atom is nullptr");
+            }
+            this->atoms.push_back(atom);
+        }
+        if (!this->atoms.empty() && this->atoms.size() >= BATCH_SIZE) {
```
andre-senna:
```suggestion
        if (this->atoms.size() >= BATCH_SIZE) {
```

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,306 @@
...
+        Utils::sleep();
+        return false;
+    }
+
+    try {
+        for (size_t i = 0; i < BATCH_SIZE; ++i) {
+            if (this->input_queue->empty()) break;
+            auto atom = (Atom*) this->input_queue->dequeue();
+            this->count++;
+            if (atom == nullptr) {
```
andre-senna: Use `nullptr` when you make comparison with `unique_ptr<type>` or `shared_ptr<type>`. 
When using raw pointers (Type*) you should use `NULL` instead.

andre-senna:
```suggestion
            if (atom == NULL) {
```

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,306 @@
...
+void BatchConsumer::send_batch(vector<Atom*> atoms, int batch_id) {
+    auto start = chrono::steady_clock::now();
+
+    LOG_DEBUG("Batch #" << batch_id << " started | size: " << atoms.size()
+                        << " | thread: " << this_thread::get_id());
+
+    try {
+        this->atomdb->add_atoms(atoms, false, true);
+
+        auto elapsed =
```
andre-senna: Use Utils::StopWatch instead of calling time functions directly.

src/db_adapter/BoundedSharedQueue.h
marcocapozzoli: You're right, I changed the code and forgot to remove that member

src/db_adapter/DatabaseAdapter.cc
```diff
@@ -0,0 +1,92 @@
...
+    }
+    LOG_DEBUG("Loaded " + to_string(this->tables_mapping.size()) + " table mappings from context file.");
+    if (!this->queries_SQL.empty()) {
+        for (size_t i = 0; i < this->queries_SQL.size(); i++) {
+            db_mapping_job.add_task_query("custom_query_" + to_string(i), this->queries_SQL[i]);
+        }
+    }
+    LOG_DEBUG("Loaded " + to_string(this->queries_SQL.size()) + " queries from query file.");
+
+    unsigned int num_threads = 8;
```
ccgsnet: Can we make this configurable?

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,308 @@
+#include "DatabaseLoader.h"
+
+#include "AtomDBSingleton.h"
+#include "BoundedSharedQueue.h"
+#include "DedicatedThread.h"
+#include "Logger.h"
+#include "PostgresWrapper.h"
+#include "Processor.h"
+
+#define LOG_LEVEL DEBUG_LEVEL
```
ccgsnet: Downgrade it to `INFO_LEVEL`

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,308 @@
+#include "DatabaseLoader.h"
+
+#include "AtomDBSingleton.h"
+#include "BoundedSharedQueue.h"
+#include "DedicatedThread.h"
+#include "Logger.h"
```
ccgsnet: You can remove this one.

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    this->current_task++;
+    this->finished = (this->current_task >= this->tasks.size());
+    return !this->finished;
+}
+
+bool DatabaseMappingJob::is_finished() const { return this->finished; }
+
+AtomPersistenceJob::AtomPersistenceJob(shared_ptr<BoundedSharedQueue> input_queue)
```
ccgsnet: Are we using this class?

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,308 @@
...
+    LOG_INFO("BatchConsumer destroyed | total_atoms: "
+             << this->total_count.load() << " | batches_dispatched: " << this->batches_dispatched.load()
+             << " | batches_completed: " << this->batches_completed.load()
+             << " | batches_failed: " << this->batches_failed.load());
+}
+
+void BatchConsumer::dispatch() {
+    if (static_cast<size_t>(this->pool.size()) >= this->max_pending_batches) {
+        this->flush_batch();
+        return;
```
ccgsnet, line 178: Don't we need to process the final remainder (batch) before returning here?

src/db_adapter/DatabaseLoader.cc
```diff
@@ -0,0 +1,308 @@
...
+
+    if (limit == 0) return;
+
+    size_t drained = 0;
+
+    while (drained < limit) {
+        if (this->input_queue->empty()) break;
+
+        void* raw_ptr = this->input_queue->dequeue();
+        queue<Atom*>* batch_queue = static_cast<queue<Atom*>*>(raw_ptr);
```
ccgsnet, line 232: What if we have `SQL2METTA` set? Wouldn't this break as it would try to cast a `string` to `Atom*` here?

src/db_adapter/DatabaseLoader.h
```diff
@@ -1,12 +1,18 @@
...
 #include "AtomDBSingleton.h"
+#include "BoundedSharedQueue.h"
 #include "DedicatedThread.h"
 #include "PostgresWrapper.h"
 #include "Processor.h"
-#include "SharedQueue.h"
+#include "processor/ThreadPool.h"
 
-#define BATCH_SIZE 2000
+#define BATCH_SIZE 100000
```
ccgsnet, line 16: Isn't `100_000` too big?
Even though I think we should be able to set it on the fly.

src/db_adapter/DatabaseMapper.cc
```diff
@@ -227,10 +235,14 @@ SQL2AtomsMapper::~SQL2AtomsMapper() {}
 
 OutputList SQL2AtomsMapper::get_output() { return this->atoms; }
 
-void SQL2AtomsMapper::clear() { this->atoms.clear(); }
+void SQL2AtomsMapper::clear() {
+    this->atoms.clear();
+    // this->unique_handles.clear();
```
ccgsnet: Leftover?

src/db_adapter/DatabaseMapper.cc
```diff
@@ -93,9 +93,14 @@ string BaseSQL2Mapper::escape_inner_quotes(string text) {
 }
 
 bool BaseSQL2Mapper::insert_handle_if_missing(const string& handle) {
```
ccgsnet: Are we using this logic?
If it consume too much memory we should discuss another approaches to try to mitigate the duplicates issues.

src/tests/cpp/db_adapter_test.cc
```diff
@@ -163,9 +166,62 @@ class PostgresWrapperTest : public ::testing::Test {
...
+        unordered_set<string> atom_handles;
+
+        while (true) {
+            if (q->empty()) break;
+            void* raw_ptr = q->dequeue();
+            std::queue<Atom*>* batch_queue = static_cast<std::queue<Atom*>*>(raw_ptr);
+
+            if (batch_queue != nullptr) {
+                while (!batch_queue->empty()) {
+                    Atom* atom = batch_queue->front();
```
ccgsnet, line 179: nit: add `delete atom;` after using it.

src/db_adapter/DatabaseLoader.cc
marcocapozzoli: No. I removed it

src/db_adapter/DatabaseLoader.h
marcocapozzoli: 100,000 is a very large number I agree, but in my tests it worked well. Both MongoDB and Mork handled the amount of data well.

src/db_adapter/DatabaseLoader.cc
marcocapozzoli: This part is only used to send what's in the accumulator and free up space in the pool.

src/db_adapter/DatabaseLoader.cc
marcocapozzoli: This implementation deals only with Atoms

ccgsnet (approved):

PR #1071, opened 2026-04-13
[AtomDB] Remove ENV VARS

ccgsnet:
This PR removes the ENV VARs from the code base (mostly regarding AtomDB).
The idea is getting those values from a give `JsonConfig` now.
Where we once used default/test values, I created a helper functions to build a `JsonConfig` object with those values.

TODO: `bus_client` is still not using the `"param"` section from the `JsonConfig`, that will be implemented in a following PR.


src/agents/context_broker/README.md
```diff
@@ -14,31 +14,6 @@ This will generate the binaries for all components in the `das/bin` directory.
 
 ## **Prerequisites**
 
-Before running the Context Broker, ensure that the following environment variables are properly exported (or set in `.env` file):
```
andre-senna, line 17: Just checking... Nothing to add to the README after these changes?

src/agents/query_engine/README.md
```diff
@@ -14,37 +14,6 @@ This will generate the binaries for all components in the `das/bin` directory.
 
 ## **Prerequisites**
 
-Before running the Query Agent, ensure that the following environment variables are properly exported:
```
andre-senna, line 17: Just checking... Nothing to add to the README after these changes?

src/atomdb/AtomDBSingleton.h
```diff
@@ -20,8 +21,8 @@ namespace atomdb {
 class AtomDBSingleton {
    public:
     ~AtomDBSingleton() {}
-    static void init(
-        atomdb_api_types::ATOMDB_TYPE atomdb_type = atomdb_api_types::ATOMDB_TYPE::REDIS_MONGODB);
+    static void init(atomdb_api_types::ATOMDB_TYPE atomdb_type,
```
andre-senna: Why do we need to pass atomdb_type separately? Isn't it in the JSON?

src/main/helpers/Helper.cc
```diff
@@ -298,4 +303,13 @@ ProcessorType Helper::processor_type_from_string(const string& type_str) {
     } else {
         return ProcessorType::UNKNOWN;
     }
-}
\ No newline at end of file
+}
+
+commons::JsonConfig Helper::default_atomdb_json_config() {
```
andre-senna: Why do we need such a default JSON? And why does it have only MongoDB configuration?
If possible, I'd prefer not having a default JSON at all, forcing the caller to always explicitly using one.

src/main/bus_client.cc
```diff
@@ -2,9 +2,11 @@
 
 #include "FitnessFunctionRegistry.h"
 #include "Helper.h"
+#include "JsonConfigParser.h"
```
andre-senna, line 6: Skipping the revision of this because I understand it will change after the discussion we had in our team sync call earlier today.

src/agents/context_broker/README.md
ccgsnet: IMO we should get rid of this README in favor of the `bus_node / bus_client` one.
But it would be better to do that in a separated PR with doc alignments only.

src/agents/query_engine/README.md
ccgsnet: Check the previous comment.

src/atomdb/AtomDBSingleton.h
ccgsnet: Done at https://github.com/singnet/das/pull/1071/commits/865e0b73c8d515b9345b996cfcc0202ac16d8737

src/main/helpers/Helper.cc
ccgsnet: Done at https://github.com/singnet/das/pull/1071/commits/865e0b73c8d515b9345b996cfcc0202ac16d8737

andre-senna (approved):

PR #1086, opened 2026-04-30
[#1009] Refactory in query variables ATTENTION_UPDATE_FLAG -> ATTENTION_UPDATE + ATTENTION_CORRELATION

andre-senna:
WIP towards #1009 

In this PR we REMOVED the BaseQueryProxy variable ATTENTION_UPDATE_FLAG and replaced it by 2 new variables ATTENTION_UPDATE and ATTENTION_CORRELATION, both typed `unsigned int`.

BaseQueryProxy is the base class for PattenrMatchingQueryProxy and QueryEvolutionProxy so both commands are affected in two ways. (1) First, the optional parameters has changed, by removing a flag and adding two new unsigned ints and (2) the tokenization of the queries has changed accordingly.

Both new variables uses only the values {0, 1, 2, 3} as in the following enumeration (defined in BaseQueryProxy.h):

```
    enum QUERY_ELEMENTS { NONE = 0, HANDLES, VARIABLES, HANDLES_AND_VARIABLES };
```


src/main/helpers/Helper.cc
```diff
@@ -25,7 +25,8 @@ string Helper::TOKENS = "tokens";
 string Helper::REQUEST = "request";
 string Helper::TIMEOUT = "timeout";
 string Helper::MAX_ANSWERS = "max-answers";
-string Helper::ATTENTION_UPDATE_FLAG = "attention-update-flag";
+string Helper::ATTENTION_UPDATE = "attention-update";
+string Helper::ATTENTION_CORRELATION = "attention-correlation";
```
ccgsnet, line 29: We must also update the "help" text on this file around these changes.
And also update/add them to `config/client.json` file.

ccgsnet (approved):

src/main/helpers/Helper.cc
andre-senna: OK. I'll merge anyway and add the requested fixes in the next PR

PR #1091, opened 2026-05-05
[#1074] Transforming the Database Adapter into an AtomDB

### LINKED ISSUE #1074: Transforming the Database Adapter into an AtomDB

marcocapozzoli:
- Added AdapterDB as a new AtomDB implementation.
- Moved the old Database Adapter pipeline into this new backend-oriented flow.
- Updated configuration to support source database credentials, mapping file paths, MeTTa export, and selectable AtomDB backends.
- Added MongoDB-based context tracking to avoid reloading the same mapped context.
- Removed the old standalone DatabaseAdapter path and adjusted build/config files accordingly.

Resolves #1074


.github/scripts/setup_bazel.sh
```diff
@@ -135,6 +135,7 @@ cd libpqxx-7.10.5
 
 cmake -S . -B build \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
```
andre-senna, line 138: Just out of curiosity, why do we need this new flag?

config/das.json
```diff
@@ -29,6 +29,59 @@
                 }
             ]
         },
+        "adapter": {
```
andre-senna: This key should be `adapterdb` just like the key in the Singleton

src/atomdb/adapterdb/AdapterDB.h
```diff
@@ -0,0 +1,124 @@
...
+    string context_id() const;
+
+    vector<string> get_mapping_file_contents() const;
+    void mongodb_setup();
+    void get_atomdb_backend();
+
+    /**
+     * Runs the full adapter pipeline (orchestrator -> persister) synchronously.
+     */
+    void sync_source_database_to_atomdb();
```
andre-senna: rename to synchronous_source_database_to_atomdb() because "sync" is ambiguous (it could also mean "synchronize" which would also make sense but would be a completely different matter).

src/atomdb/adapterdb/AdapterDB.h
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Produces a string that identifies this mapping context (a hash)
+     * Used to decide whether the backend already holds this data.
+     */
+    string context_id() const;
+
+    vector<string> get_mapping_file_contents() const;
+    void mongodb_setup();
+    void get_atomdb_backend();
```
andre-senna: A "get" method that returns void is weird. Rename properly.

src/atomdb/adapterdb/AdapterDB.cc
```diff
@@ -0,0 +1,377 @@
...
+    } else if (atomdb_backend_type == "redismongodb") {
+        this->atomdb_backend = shared_ptr<AtomDB>(new RedisMongoDB("", false, atomdb_backend_config));
+    } else if (atomdb_backend_type == "remotedb") {
+        this->atomdb_backend = shared_ptr<AtomDB>(new RemoteAtomDB(atomdb_backend_config));
+    } else {
+        Utils::error("Invalid AtomDB type: " + atomdb_backend_type);
+    }
+}
+
+bool AdapterDB::is_loaded(const string& context_id) const {
```
andre-senna: This method has a lot of collateral effects considering it's called is_loaded(). Can you please explain what exactly is happening here? Should you rename it?

src/atomdb/adapterdb/AdapterDB.cc
```diff
@@ -0,0 +1,377 @@
...
+}
+
+bool AdapterDB::needs_sync() const {
+    Utils::error("needs_sync() is not implemented yet.");
+    return false;
+}
+
+bool AdapterDB::allow_nested_indexing() { return this->atomdb_backend->allow_nested_indexing(); }
+
+shared_ptr<Atom> AdapterDB::get_atom(const string& handle) {
```
andre-senna, line 66: All the following `get` methods (methods that use the atomdb in the backend to retrieve data) should check if all the mapping/loading process has already been run. Since you'll need to make the check in all these methods, I suggest you to create a method like this:

```
bool check_atomdb_backend() {
    // if everything is already mapped and loaded return true, else return false
}
```

I made a comment elsewhere about your method `is_loaded()`. one of the reasons I made this comment is because you may want to reuse it to implement `check_atomdb_backend()` but I'm worried about the heavy collateral effects you put in `is_loaded()`. Perhaps an alternative would be to split `is_loaded()` in two methods, one with the collateral effects and another with just the check itself.

andre-senna (changes requested):

.github/scripts/setup_bazel.sh
marcocapozzoli: By default, libpqxx compiles as a static library (.a). And we are trying to embed it inside our dynamic library (hyperon_das.so). I saw that Linux requires the static code to be compiled in a "flexible" way (Position Independent Code).

Without this flag, the build crashes with this error:

```
ERROR: /opt/das/src/BUILD:24:18: Linking hyperon_das.so failed: (Exit 1): gcc failed: error executing CppLink command (from cc_shared_library rule target //:das) /usr/bin/gcc @bazel-out/k8-fastbuild/bin/hyperon_das.so-0.params
/usr/bin/ld.gold: error: /usr/local/lib/libpqxx.a(connection.cxx.o): requires dynamic R_X86_64_PC32 reloc against '_ZN4pqxx18quiet_errorhandlerclEPKc' which may overflow at runtime; recompile with -fPIC
/usr/bin/ld.gold: error: /usr/local/lib/libpqxx.a(connection.cxx.o): requires dynamic R_X86_64_PC32 reloc against '_ZZNSt19_Sp_make_shared_tag5_S_tiEvE5__tag' which may overflow at runtime; recompile with -fPIC
```

src/atomdb/adapterdb/AdapterDB.cc
marcocapozzoli: Sure. It returns true if the context is already loaded in MongoDB, I mean, it checks for existence and creates a record if it doesn't exist.

You're right, the method doesn't just check...

I was thinking maybe of this name `ensure_context_mapping_exists`.

src/db_adapter/DatabaseLoader.h
```diff
@@ -70,12 +73,13 @@ class MultiThreadAtomPersister {
 
    private:
     shared_ptr<BoundedSharedQueue> input_queue;
-    shared_ptr<AtomDB> atomdb;
-    shared_ptr<MettaFileWriter> metta_writer;
     ThreadPool& pool;
+    shared_ptr<AtomDB> atomdb;
+    ;
```
ccgsnet: leftover

andre-senna (approved):

PR #1117, opened 2026-05-20
[#1009] Miscellaneous fixes and small changes in the context of #1009

andre-senna:
WIP towards #1009 

The PR is a little messy because it's a lot of scattered changes in many places, all required during the fine tuning in the context of #1009 

An important change regarding Rust integration is that one of the Query Evolution parameters has been changed.

`correlation_mappings` used to be a `vector<pair<QueryAnswerElement, QueryAnswerElement>>` and now it's a `vector<vector<pair<QueryAnswerElement, QueryAnswerElement>>>`. So the tokenization of the Query Evolution proxy (request) changed a bit. To see an example, look at the diff in `query_evolution_test`.


src/attention_broker/HebbianNetwork.h
```diff
@@ -74,10 +83,33 @@ class HebbianNetwork : public Serializable {
...
+                    case GREATEST:
+                        for (auto determiner : this->determiners) {
+                            ImportanceType determiner_importance = determiner->get_importance();
+                            if (determiner_importance > answer) {
+                                answer = determiner_importance;
+                            }
+                        }
+                        break;
+                    case MULTIPLICATION_BASE_ONE:
+                        answer = 1;
```
ccgsnet, line 97: Aren't we missing a `break;` here?

ccgsnet (approved):

src/attention_broker/HebbianNetwork.h
andre-senna: Actually not. It's missing on purpose meaning that if `case MULTIPLICATION_BASE_ONE` matches, all the code below is supposed to be executed:

```
                        answer = 1;
                        for (auto determiner : this->determiners) {
                            ImportanceType determiner_importance = determiner->get_importance();
                            answer *= determiner_importance;
                        }
                        break;
```

The idea here is that both cases are supposed to multiply all the determiner importances. But in the case of `MULTIPLICATION_BASE_ONE`, multiplication starts with `1` while in the case of `MULTIPLICATION_BASE_HANDLE`, multiplication starts with the importance value of the base element, as initialized before above the switch in `ImportanceType answer = this->importance;`.

PR #1121, opened 2026-05-26
[#1102] Implementation of the MORK type AdapterDB 2

marcocapozzoli:
Implementation of MorkMappingStrategy and Metta2AtomsMapper


src/db_adapter/non_sql/mork/MorkWrapper.cc
andre-senna, line 41: I suppose this call maps the entire Mork Server contents. So you are keeping an extra copy of the atoms vector unecessarily. The container should be passed by reference to map() and be populated there directly.

This comment is coupled with another one regarding line 52 below.

You are crating yet another container (batch_queue) and make yet another copy of everything again.

I'm wondering if you cold create only batch_queue and pass its reference to be populated in mapper->map().

src/db_adapter/non_sql/Metta2AtomsMapper.cc
```diff
@@ -1,22 +1,83 @@
...
+#if LOG_LEVEL >= DEBUG_LEVEL
+    LOG_DEBUG("The expression " << metta_expression.expression
+                                << " has been mapped to the following atoms:");
+
+    for (const auto& atom : this->atoms) {
+        LOG_DEBUG("-> " << atom->to_string());
+    }
+#endif
+
+    return this->atoms;
```
andre-senna, line 50: Regarding my comments about src/db_adapter/non_sql/mork/MorkWrapper.cc

Here you keep yet another copy of all atoms.

PR #1124, opened 2026-05-27
Refactor DatabaseMapper to write directly into output queues

marcocapozzoli:

PR #1125, opened 2026-05-27
[#1102] Implementation of the MORK type AdapterDB 2

marcocapozzoli:
- Implementesd Metta2AtomsMapper, MorkWrapper and MorkMappingStrategy
- Implemented loading of .metta query files
- Improved and added tests

### DIFF
--- a/src/db_adapter/non_sql/mork/MorkWrapper.cc
+++ b/src/db_adapter/non_sql/mork/MorkWrapper.cc
@@ -17,9 +17,7 @@ using namespace db_adapter;
 MorkWrapper::MorkWrapper(shared_ptr<MorkConnection> conn,
                          shared_ptr<BoundedSharedQueue> output_queue,
                          MAPPER_TYPE mapper_type)
-    : DatabaseWrapper(conn, MapperFactory::create(mapper_type)),
-      conn(conn),
-      output_queue(output_queue) {}
+    : DatabaseWrapper(conn, MapperFactory::create(mapper_type)), output_queue(output_queue) {}
 
 MorkWrapper::~MorkWrapper() {}
 
--- a/src/db_adapter/non_sql/mork/MorkWrapper.cc
+++ b/src/db_adapter/non_sql/mork/MorkWrapper.cc
@@ -15,17 +15,17 @@ using namespace db_adapter;
 //  Construction / destruction
 // ==============================
 
-MorkWrapper::MorkWrapper(MorkConnection& conn,
+MorkWrapper::MorkWrapper(shared_ptr<MorkConnection> conn,
                          shared_ptr<BoundedSharedQueue> output_queue,
                          MAPPER_TYPE mapper_type)
-    : DatabaseWrapper(conn, MapperFactory::create(mapper_type)),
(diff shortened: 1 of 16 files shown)


src/db_adapter/non_sql/mork/MorkWrapper.cc
```diff
@@ -34,27 +34,16 @@ void MorkWrapper::map(const string& metta_query) {
     for (const auto& expr : metta_expressions) {
         LOG_DEBUG("Mapping MORK expression: " << expr);
 
-        vector<shared_ptr<Atom>> atoms;
+        std::queue<shared_ptr<Atom>>* batch_queue = new std::queue<shared_ptr<Atom>>();
```
andre-senna, line 37: I didn't see where `batch_queue` is being deleted

src/db_adapter/non_sql/mork/MorkWrapper.cc
```diff
@@ -15,17 +15,17 @@ using namespace db_adapter;
 //  Construction / destruction
 // ==============================
 
-MorkWrapper::MorkWrapper(MorkConnection& conn,
+MorkWrapper::MorkWrapper(shared_ptr<MorkConnection> conn,
                          shared_ptr<BoundedSharedQueue> output_queue,
                          MAPPER_TYPE mapper_type)
-    : DatabaseWrapper(conn, MapperFactory::create(mapper_type)),
+    : DatabaseWrapper(*conn, MapperFactory::create(mapper_type)),
```
andre-senna: This very error prone. Inside DatabaseWrapper you are storing a reference to an object attached to a shared_ptr. but this reference is not known by the shared_ptr which means that when the count in shared_ptr reaches ZERO (i.e. when all the shared_ptr which have copied the same obj is deleted) the object itself is deleted turning the reference inside DatabaseWrapper bogus.

I suggest passing (and storing inside DatabaseWrapper the shared_ptr itself instead of a reference to the object.

src/db_adapter/non_sql/mork/MorkWrapper.cc
```diff
@@ -15,17 +15,17 @@ using namespace db_adapter;
...
       conn(conn),
```
andre-senna: This is even weirder. The superclass already store a connection object. And here you have the subclass storing it again. This is a design flaw.

src/db_adapter/non_sql/mork/MorkWrapper.h
```diff
@@ -24,7 +24,7 @@ class MorkWrapper : public DatabaseWrapper {
 
    private:
     mutex api_mutex;
-    MorkConnection& conn;
+    shared_ptr<MorkConnection> conn;
```
andre-senna: See comment above. The superclass already has an object of this type.

andre-senna (changes requested):

src/db_adapter/non_sql/mork/MorkWrapper.cc
marcocapozzoli: I replied here: https://github.com/singnet/das/pull/1124#discussion_r3317899902

andre-senna (approved):

PR #1129, opened 2026-05-29
[#1102] Implementation of the MORK type AdapterDB 3

### LINKED ISSUE #1102: Implement an Adapter to MORK

marcocapozzoli:
Add MORK support to AdapterDB

Resolves #1102

### DIFF
--- a/src/db_adapter/AtomPersister.cc
+++ b/src/db_adapter/AtomPersister.cc
@@ -189,10 +189,13 @@ void AtomPersister::send_batch(vector<shared_ptr<Atom>> atoms,
 
         timer_success.stop();
 
-        int new_total =
-            this->total_count.fetch_add(static_cast<int>(atoms.size())) + static_cast<int>(atoms.size());
+        this->total_count.fetch_add(static_cast<int>(atoms.size())) + static_cast<int>(atoms.size());
 
         this->batches_completed.fetch_add(1);
+
+#if LOG_LEVEL >= DEBUG_LEVEL
+        int new_total = this->total_count.load();
+#endif
 
         LOG_DEBUG("Batch #" << batch_id << " completed | size: " << atoms.size()
                             << " | time: " << timer_success.milliseconds() << "ms"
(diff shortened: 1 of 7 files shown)


src/db_adapter/AtomPersister.cc
```diff
@@ -188,10 +188,10 @@ void AtomPersister::send_batch(vector<shared_ptr<Atom>> atoms,
         }
 
         timer_success.stop();
-#if LOG_LEVEL >= DEBUG_LEVEL
```
andre-senna, line 191: Why removing this? I understand `new_total` is used only inside the `LOG_DEBUG` below. Thus, if you remove this we'll have a compiling warning (complaining about `new_total` being defined but never used) whenever LOG_LEVEL if INFO or below.

src/db_adapter/ContextLoader.cc
```diff
@@ -59,7 +59,11 @@ vector<string> ContextLoader::load_metta_queries(const string& file_path) {
     int parentheses_depth = 0;
 
     while (getline(file, line)) {
-        if (line.empty()) continue;
+        bool is_comment = line.size() >= 2 && line[0] == ';';
```
andre-senna:
```suggestion
        bool is_comment = (line.size() >= 2) && (line[0] == ';');
```

andre-senna: I suggest adding () as above. Not required, true, but it make the code less error prone. 

A more important comment though is the first comparison `line.size() >= 2`. Is this really what you meant? By this definition, a line of size 1 with only a `;` is not a comment.

andre-senna (approved):

src/db_adapter/AtomPersister.cc
marcocapozzoli: Yes, I use `new_total` in the `LOG_DEBUG` below. But I also use `total_count` elsewhere. Anyway, I think I did better. I can use `total_count` however I want, and I also added `#if` again in `new_total`.

andre-senna (approved):

PR #1134, opened 2026-06-01
[#1133] Add thread id to non-INFO log messages

andre-senna:
```
2026-06-01 20:17:23 | ERROR | 6977 agents/BaseProxy.cc:92 raise_error | Exception thrown in command processor.
                              ----
                            Thread id
```


src/tests/cpp/utils_test.cc
```diff
@@ -79,4 +79,5 @@ TEST(UtilsTest, stack_trace) {
     EXPECT_THROW(f3(), runtime_error);
     EXPECT_THROW(f2(), runtime_error);
     EXPECT_THROW(f1(), runtime_error);
+    // FAIL();
```
ccgsnet, line 82: Leftover?

ccgsnet (approved):

PR #1152, opened 2026-06-18
Add sanitization for SQL column values

marcocapozzoli:
## Summary by CodeRabbit

* **New Features**
  * Added input data validation with UTF-8 character encoding verification, maximum size enforcement, and automatic whitespace normalization to ensure data integrity and consistency across the database.

* **Tests**
  * Introduced comprehensive unit tests for data sanitization behavior, validating edge cases and various input types.


src/db_adapter/sql/postgres/PostgresWrapper.cc
```diff
@@ -504,14 +504,10 @@ SqlRow PostgresWrapper::build_sql_row(const pqxx::row& row, const Table& table,
 
         string value = field.c_str();
 
-        if (value.empty() || value.size() > MAX_VALUE_SIZE) {
-            continue;
-        }
-
-        Utils::replace_all(value, "\n", " ");
-        Utils::replace_all(value, "\t", " ");
         value = Utils::trim(value);
```
andre-senna: Why did you keep this outside of `sanitize_value()`?

src/db_adapter/sql/postgres/PostgresWrapper.cc
```diff
@@ -504,14 +504,10 @@ SqlRow PostgresWrapper::build_sql_row(const pqxx::row& row, const Table& table,
 
         string value = field.c_str();
```
andre-senna: This line is suspicious.
I don't know for sure the type of `field` since you declared it `auto` (btw, try to avoid using `auto` when readability of code is not in check), but it seems like it's a `string`. So this line is getting the `char*` out of a string just to rebuild a `string` object with it. Why not making simply:

```
string value = field;
```

instead? Or, even better... Why not using `field` directly? `field` is already a copy string of `row[I]`.

ccgsnet:
@coderabbitai review

src/db_adapter/sql/postgres/PostgresWrapper.cc
marcocapozzoli: In `auto field = row[i];`, `field` is not a string; it is of type `pqxx::field`. Therefore, to convert it to a string, as shown in the documentation itself, you do it as follows:

`string value = field.c_str();`

or

`string value = field.as<string>();`

src/db_adapter/sql/postgres/PostgresWrapper.cc
marcocapozzoli: I missed that.

src/db_adapter/sql/SqlWrapper.cc
```diff
@@ -1,6 +1,22 @@
...
+
 SQLWrapper::SQLWrapper(shared_ptr<DatabaseConnection> db_conn, shared_ptr<DatabaseMapper> mapper)
-    : DatabaseWrapper(db_conn, mapper) {}
\ No newline at end of file
+    : DatabaseWrapper(db_conn, mapper) {}
+
+bool SQLWrapper::sanitize_value(string& value) {
+    value = Utils::trim(value);
+
+    if (value.empty() || value.size() > MAX_VALUE_SIZE || !utf8::is_valid(value)) return false;
```
andre-senna, line 13: This call should be made before trim()

andre-senna (approved):

PR #1156, opened 2026-06-19
[#1146] Adds the initial HTTP server infrastructure for the command router.

marcocapozzoli:
The server boots when the `COMMAND_ROUTER` service starts and exposes a health-check endpoint.

This is the minimum working version to validate the setup.

Execution and WebSocket endpoints will be added in follow-up PRs.

- Add `CommandRouterHttpAPI` (httplib server on a dedicated thread)
- Add `CommandRouterHttpAPISingleton` for initialization from config
- Wire startup in `bus_node` using `http_api_host` and `http_api_port`
- Integrate lifecycle with `BusCommandRouterProcessor`
- Implement `GET /ping` → `PONG!`


src/tests/cpp/bus_command_router_test.cc
```diff
@@ -17,6 +18,12 @@ class BusCommandRouterTestEnvironment : public ::testing::Environment {
    public:
     void SetUp() override {
         AtomDBSingleton::init(test_atomdb_json_config());
+
+        auto json = nlohmann::json();
+        json["http_api_host"] = "localhost";
+        json["http_api_port"] = 19002;
+        CommandRouterHttpAPISingleton::init(commons::JsonConfig(json));
```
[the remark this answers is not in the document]
marcocapozzoli: This is fine since it's just a test. The team uses it (hardcoded) in other parts of the code.

config/das.json
```diff
@@ -201,7 +201,9 @@
         },
         "command_router": {
             "endpoint": "localhost:40008",
-            "ports_range": "48000:48999"
+            "ports_range": "48000:48999",
+            "http_api_host": "localhost",
```
ccgsnet: Can we use `"localhost:40009"` as we do for all other endpoints here?

src/agents/atomdb_broker/AtomDBProxy.h
```diff
@@ -177,7 +177,7 @@ class AtomDBProxy : public BaseProxy {
     shared_ptr<SharedQueue> processing_queue = nullptr;
     size_t pending_atoms_count = 0;
     thread processing_thread;
-    shared_ptr<ThreadPool> thread_pool;
+    shared_ptr<::ThreadPool> thread_pool;
```
ccgsnet: Leftover?

src/agents/link_creation_agent/LinkCreationService.h
```diff
@@ -63,7 +63,7 @@ class LinkCreationService
     ~LinkCreationService();
 
    private:
-    ThreadPool thread_pool;
+    ::ThreadPool thread_pool;
```
ccgsnet: Leftover?

src/agents/command_router/CommandRouterHttpAPISingleton.cc
```diff
@@ -0,0 +1,59 @@
...
+        auto thread_pool_executor =
+            make_shared<ThreadPool>("http_api_thread_pool_executor", num_threads);
+
+        HTTP_API = make_shared<CommandRouterHttpAPI>(host, port, thread_pool_executor);
+
+        auto http_api_thread = make_shared<DedicatedThread>("http_api_thread", HTTP_API.get());
+
+        try {
+            CommandRouterHttpAPI::initialize(HTTP_API, {thread_pool_executor, http_api_thread});
+            Utils::sleep(200);  // time to bind the port
```
ccgsnet: Isn't there a way to check if the bind succeeded here?

src/agents/command_router/BusCommandRouterProcessor.cc
```diff
@@ -25,9 +26,15 @@ const string CONTEXT_KEY = "context";
...
 BusCommandRouterProcessor::BusCommandRouterProcessor(shared_ptr<ServiceBus> service_bus)
-    : BusCommandProcessor({ServiceBus::BUS_COMMAND_ROUTER}), service_bus(service_bus) {}
+    : BusCommandProcessor({ServiceBus::BUS_COMMAND_ROUTER}), service_bus(service_bus) {
+    this->http_api = CommandRouterHttpAPISingleton::get_instance();
+}
 
-BusCommandRouterProcessor::~BusCommandRouterProcessor() {}
+BusCommandRouterProcessor::~BusCommandRouterProcessor() {
+    if (this->http_api != nullptr) {
+        this->http_api->stop();
```
ccgsnet: We are stopping it here but that object comes from the `CommandRouterHttpAPISingleton`.
We should stop it there so its lifecycle stays in the singleton.

src/agents/atomdb_broker/AtomDBProxy.h
marcocapozzoli: ::ThreadPool was a workaround for the name clash between commons/ThreadPool.h and processor::ThreadPool after adding the command_router. I'll remove using namespace processor from the header and use processor::ThreadPool explicitly instead.

andre-senna: So use processor::ThreadPool  instead, please, to make it clear which one you are using. Empty namespace names like ::f() are meant to be used when you have a symbol in your code that clashes with some of a library you are using. To disambiguate between a symbol of two different libraries, the proper way to go is to use the proper namespace.

src/agents/link_creation_agent/LinkCreationService.h
marcocapozzoli: https://github.com/singnet/das/pull/1156#discussion_r3463768008

src/agents/command_router/CommandRouterHttpAPISingleton.cc
marcocapozzoli: I replaced it with httplib's recommended split:

- setup() (init thread): bind_to_port() runs synchronously and fails fast if the port is unavailable.

- thread_one_step() (DedicatedThread): listen_after_bind() starts the accept loop and blocks there until stop().

So initialize() only returns after a successful bind, without a fixed delay. If the port is already in use, startup fails immediately with a clear error.

src/agents/command_router/BusCommandRouterProcessor.cc
marcocapozzoli: Removed the HTTP API reference from BusCommandRouterProcessor entirely. The processor never used it, lifecycle is owned by CommandRouterHttpAPISingleton::init() in bus_node, which runs before the processor is created. The bus processor and HTTP server are independent components.

src/tests/cpp/command_router_http_api_test.cc
```diff
@@ -0,0 +1,137 @@
...
+    EXPECT_EQ(CommandRouterHttpAPISingleton::get_instance().get(), api2.get());
+    EXPECT_NE(CommandRouterHttpAPISingleton::get_instance().get(), api1.get());
+}
+
+TEST_F(CommandRouterHttpAPISingletonTest, calling_init_after_provide_throws_already_initialized) {
+    auto pool = make_shared<ThreadPool>("double_init_pool", 1);
+    auto api = make_shared<CommandRouterHttpAPI>("localhost", 19005, pool);
+    CommandRouterHttpAPISingleton::provide(api);
+
+    json raw = {{"http_api_host", "localhost"}, {"http_api_port", 19005}};
```
ccgsnet, line 130: I think we should update these keys to reflect the new one (even knowing that the test does not deal with it)

ccgsnet (approved):

src/tests/cpp/command_router_http_api_test.cc
marcocapozzoli: The new PR update theses keys: https://github.com/singnet/das/pull/1165/changes#diff-b017fff98e4d9c8ab61e69bf88d3c23b620a98ed958004805d6b1fc126994abdR301

PR #1160, opened 2026-06-23
[#1157] Added a way to control Attention Focus strictness through a tunnable parameter

andre-senna:
Resolves @1157

In this PR we add a step inside the processing of LinkTemplates (the basic query elements in the leafs of PM query trees) to prune the results according to a strictness [0..1] parameter. Pruning is performed based on atom STI. The idea is to prevent low-STI-valued atoms from advancing through the query tree early in the query processing to prevent unnecessary use of RAM and processor time. This behavior can be understood (in cognitive terms) like we are making attention focus more or less strict.

This strictness is controlled by a new PM Query parameter ATTENTION_FOCUS_STRICTNESS which is a [0..1] value and can be set independently by query. It means that the AF strictness for different queries are independent i.e. we can eventually process multiple queries simultaneously each with its own different  ATTENTION_FOCUS_STRICTNESS value.

Attention Focus is pruned using this parameter but the algorithm actually used to make the pruning is also configurable. We could easily add different policies. Currently implemented policy is simply interpreting ATTENTION_FOCUS_STRICTNESS as a percentage and just prune x% of the atoms which would be answered by the LinkTemplate (based in their STI).

The boundary values for ATTENTION_FOCUS_STRICTNESS 0.0 and 0.1 have special meaning. 0.0 means that no pruning will be done and Attention Focus will encompass all the atoms returned in the LinkTemplate. 1.0 means that only one atom (the one with larger STI) will be returned. If more the one atom has the same largest STI, the winner is randomly chosen (it is not actually "chosen"... But a random choice is guaranteed by the way LinkTemplate sort the atoms by their STI).

It's important to notice that  ATTENTION_FOCUS_STRICTNESS does not prune results at the end of the query tree but only in its beginning, i.e. in the LinkTemplates which are their leaves. It means, for example, that setting ATTENTION_FOCUS_STRICTNESS = 1.0 doesn't mean that we should expect only one answer to the query.

Summary of changes:

* ATTENTION_FOCUS_STRICTNESS was actually added in a previous PR: https://github.com/singnet/das/pull/1158
* PatternMatchingQueryProxy was changed to add the new parameter string tag
* PatternMatchingQueryProcessor and MettaParserActions were changed to consider this parameter when building LinkTemplate
* LinkTemplate was changed to expect this new parameter and use it as an additional processing layer of the fetched results during a query. Now QueryAnswers are fetched -> matched -> reported. Before this PR, all matched QueryAnswers were reported, now, only the answers considered to be IN the AttentionFocus are actually reported. In this context, "reported" means to be passed to the next elements in the query tree.
* Old test cases has been updated accordingly to cope with the new parameter and new test cases has been added to cover the new PM query parameter.


src/agents/query_engine/query_element/LinkTemplate.cc
```diff
@@ -166,6 +202,10 @@ void LinkTemplate::processor_method(shared_ptr<StoppableThread> monitor) {
             compute_importance(tagged_handles);
         }
     }
+    vector<AttentionFocusRecord> attention_focus_candidates;
+    if ((this->attention_focus_strictness != 0.0) || (this->attention_focus_strictness != 1.0)) {
```
ccgsnet: Is this condition correct? It always returns `true`. I think that should be a `&&`.

src/agents/query_engine/query_element/LinkTemplate.cc
```diff
@@ -129,6 +140,30 @@ void LinkTemplate::compute_importance(vector<pair<char*, float>>& handles) {
               });
 }
 
+unsigned int LinkTemplate::report_attention_focus_by_percentage(
+    vector<AttentionFocusRecord>& attention_focus_candidates) {
+    unsigned int limit = lround(this->attention_focus_strictness * attention_focus_candidates.size());
```
ccgsnet, line 145: `limit` can be `0` for low strictness and small candidates size (eg `lround(0.1 * 4) = 0`).
Maybe we should fallback to `1` when that happens.

ccgsnet (approved):

src/agents/query_engine/query_element/LinkTemplate.cc
andre-senna: Fixed

src/agents/query_engine/query_element/LinkTemplate.cc
andre-senna: fixed

PR #1168, opened 2026-06-26
[#1146] Route HTTP commands through BusCommandRouterProcessor in-process

marcocapozzoli:
This PR routes Command Router HTTP executions through BusCommandRouterProcessor in-process.

- Adds `dispatch_http_command()` to set up caller/processor proxy peers locally and run commands through the existing router processor
- Introduces `BusCommandRouterProxyStreamPoller` to stream chunked results for get, set, query, and evolution
- Wires CommandRouterHttpAPI to dispatch via the processor and publish streamed events over HTTP/WebSocket
- Requires a BusCommandRouterProcessor at startup in CommandRouterHttpAPISingleton


marcocapozzoli:
@coderabbitai full review

marcocapozzoli:
@coderabbitai full review

src/agents/command_router/http_api/CommandRouterHttpAPI.cc
```diff
@@ -280,70 +289,88 @@ shared_ptr<CommandExecution> CommandRouterHttpAPI::find_execution(const string&
...
-    // POC code to simulate event streaming.
+    if (this->router_processor == nullptr) {
+        exec->mark_error("Command router processor is not configured for HTTP execution");
+        return;
+    }
 
-    const int TOTAL = 1000;
-    const int CHUNK_SIZE = 10;
-    auto started_at = Utils::get_current_time_millis();
+    const auto started_at = Utils::get_current_time_millis();
```
andre-senna: Use Utils::StopWatch to measure time. 

StopWatch timer;

After calling timer.start(), you can use timer.milliseconds() to get the elapsed time in milliseconds if this is really required. If you are measuring time just to print LOG messages, using timer.str_time() will print a  more friendly string when the measured time is larger than 1 second.

src/agents/command_router/http_api/CommandRouterHttpAPI.cc
```diff
@@ -280,70 +289,88 @@ shared_ptr<CommandExecution> CommandRouterHttpAPI::find_execution(const string&
...
+        exec->mark_aborted_unless_terminal();
+        return;
     }
 
-    lock_guard<mutex> semaphore(this->executions_mtx);
+    exec->mark_running();
+
+    this->run_execution_inner(exec);
+
+    lock_guard<mutex> lock(this->executions_mtx);
```
andre-senna: This is in the "global" scope of the function run_execution() so the mutex is locked prior to the first command of the function. Thus, this line should be moved to the top of the function for clarity.

That being said, I'm not sure you are having the expected behavior of the locking dynamics here.  Actually the locking dynamics is particularly confuse in this method, please explain it in a higher level explanation so we can assess if the actually used synchronization methods are performing the desired actions.

marcocapozzoli: The mutex is used in two sections, not for the whole function:

1. At the start: wait for a slot, update counters.
```
{
    unique_lock<mutex> lock(this->executions_mtx);
    ...
}
```
2. At the end: release the slot and wake the next worker.
`lock_guard<mutex> lock(this->executions_mtx);`

Everything between those two lock blocks runs unlocked, so we can have multiple executions in flight (up to max_concurrent_executions).

Moving that lock to the top would keep the mutex held during all of `run_execution_inner()`, and that would pretty much kill concurrency.

Regarding clarity, I can wrap the final `lock` block in `{...}` to match the pattern of the first block. The behavior is the same, but I think it makes it easier to read.

andre-senna: `unique_lock<mutex> lock(this->executions_mtx);` is not a command but a declaration of a variable named `lock` which is valid in the entire scope. In this case, in the scope of the function. It's confuse to declare it in the middle of a block as you did and I'm not sure C++ have assurance that the lock will only be acquired when the flow reaches the declaration line rather than when the activation record is instantiated in the beginning of the function. Gemini said it has but this can be specific for some compilers.

Anyway, please enclose it in a explicit scope using `{` `}` to improve clarity.

andre-senna: After better reading the code, it seems like you actually want to put the scope like this:

```
    {
        lock_guard<mutex> lock(this->executions_mtx);
        this->running_executions--;
    }
    this->execution_slots_cv.notify_one();
```

Otherwise you'd awake whoever is waiting on the cv just to put it to sleep again because this scope still have the lock.

ccgsnet (approved):

andre-senna (approved):

PR #1169, opened 2026-06-29
[#BUGFIX] Fix flaky adapterdb tests with MORK retry logic and bump MORK image to 1.1.0

marcocapozzoli:
Fixes intermittent adapterdb_test failures in CI caused by HTTP 401 from the MORK server when tests run in parallel and contend for the same path.

- Add retry with exponential backoff to MorkClient for transient 401/409 responses
- Update MORK Docker images from 1.0.4/1.0.5 to 1.1.0 across test and setup scripts

### DIFF
--- a/src/atomdb/morkdb/MorkDB.cc
+++ b/src/atomdb/morkdb/MorkDB.cc
@@ -24,8 +24,6 @@ using namespace metta;
 namespace {
 
 bool is_transient_mork_http_status(int status) {
-    // MORK returns 401 when a path is temporarily locked by a concurrent request, and 409 for
-    // PathForbiddenTemporary. Both are expected under parallel test load and should be retried.
     return status == httplib::StatusCode::Unauthorized_401 ||
            status == httplib::StatusCode::Conflict_409;
 }
@@ -66,7 +64,7 @@ string MorkClient::clear(const string& pattern) {
 
 httplib::Result MorkClient::send_request(const string& method, const string& path, const string& data) {
     const unsigned int max_attempts = 16;
-    const unsigned int initial_delay_ms = 5;  // 0.005s, matching mork_loader.py
+    const unsigned int initial_delay_ms = 5;
     const double backoff = 2.0;
 
     for (unsigned int attempt = 0; attempt < max_attempts; ++attempt) {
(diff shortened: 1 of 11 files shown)


src/scripts/mork_server.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eoux pipefail
 
-IMAGE_NAME="trueagi/das:mork-server-1.0.0"
+IMAGE_NAME="trueagi/das:mork-server-1.0.4"
```
ccgsnet: The latest one is [mork-server-1.1.0](https://hub.docker.com/layers/trueagi/das/mork-server-1.1.0/images/sha256-4513f0cc88a9b7b808c19e68ac0f7aeb3725e89607d8dd4f965570f9cadb6fd2)

src/atomdb/morkdb/MorkDB.cc
```diff
@@ -54,25 +65,43 @@ string MorkClient::clear(const string& pattern) {
...
+            return res;
+        }
+
+        if (is_transient_mork_http_status(res->status) && attempt + 1 < max_attempts) {
+            LOG_DEBUG("MORK transient HTTP " << res->status << " at " << path << ", retrying (attempt "
+                                             << attempt + 1 << "/" << max_attempts << ")");
+            Utils::sleep(delay_ms);
+            delay_ms *= 2;
+            continue;
         }
```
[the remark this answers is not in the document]
marcocapozzoli: I modified the code to make it similar to the Python client in `/src/scripts/mork_loader.py`.

src/atomdb/morkdb/MorkDB.cc
```diff
@@ -54,25 +63,55 @@ string MorkClient::clear(const string& pattern) {
...
+                            std::to_string(max_attempts) + " retries, last status " +
+                            std::to_string(res->status));
+            }
+
+            unsigned int delay_ms = initial_delay_ms;
+            for (unsigned int i = 0; i < attempt; ++i) {
+                delay_ms = static_cast<unsigned int>(delay_ms * backoff);
+            }
+
+            LOG_DEBUG("MORK transient HTTP " << res->status << " at " << path << ", retrying (attempt "
```
ccgsnet: Lets use `LOG_ERROR` here so we can easily spot the issue if something like this pops up during the execution of a real program.

ccgsnet (approved):

PR #1189, opened 2026-07-22
[Toy] Fix `RemoteDB` concurrent logic

ccgsnet:
## Summary

Makes RemoteAtomDB safe under concurrent pattern queries (e.g. `run_toy.sh` AndNot / multiple LinkTemplates), which previously segfaulted (exit 139) due to races on the peer cache.

### InMemoryDB thread-safety
- Add an internal `recursive_mutex` and take it on every public API method.
- Mutex is recursive because public methods re-enter each other (e.g. `delete_link` → `delete_atom`, re-index visitor → `add_pattern`).
- Callers can share one `InMemoryDB` across threads without external locking.

### RemoteAtomDBPeer locking (kept simple)
- Cache atom data no longer needs peer-level locking; `InMemoryDB` owns that.
- `peer_mutex_` only guards peer bookkeeping: `cache_` pointer swap, `fetched_link_templates_`, and `staged_handles_`.
- Reads use a short-lived `cache()` snapshot (`shared_ptr`); writes hold `peer_mutex_` so staged handles stay consistent with the cache generation.
- `release_cache` never drops staged writes, even on a “clear without persistence” path.

## Test plan
- [x] `//tests/cpp:inmemorydb_test`
- [x] `//tests/cpp:remote_atomdb_test`
- [ ] `src/tests/scripts/run_toy.sh` with RemoteDB peers (concurrent LinkTemplates)


src/tests/main/evaluation_evolution.cc
```diff
@@ -292,6 +293,14 @@ static shared_ptr<PatternMatchingQueryProxy> issue_context_creation_query(
     return proxy;
 }
 
+static void flush_remote_link_template_cache(bool force = false) {
+    if (auto remote_db = dynamic_pointer_cast<RemoteAtomDB>(db); remote_db != nullptr) {
```
andre-senna:
```suggestion
    if (auto remote_db = dynamic_pointer_cast<RemoteAtomDB>(db), remote_db != nullptr) {
```

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -62,6 +62,25 @@ struct ReIndexData {
...
+        return nullptr;
+    }
+
+    if (atom->arity() == 0) {
+        auto node = dynamic_cast<const Node*>(atom);
+        return (node != nullptr) ? make_shared<Node>(*node) : nullptr;
+    }
+
+    auto link = dynamic_cast<const Link*>(atom);
+    return (link != nullptr) ? make_shared<Link>(*link) : nullptr;
```
andre-senna:
```suggestion
    auto link = dynamic_cast<const Link*>(atom);
    if (link != nullptr) {
        return make_shared<Link>(*link);
    }
    auto node = dynamic_cast<const Node*>(atom);
    if (node != nullptr) {
        return make_shared<Node>(*node);
    }
    return nullptr;
```

config/das.json
```diff
@@ -1,6 +1,6 @@
 {
     "atomdb": {
-        "type": "redismongodb",
+        "type": "remotedb",
```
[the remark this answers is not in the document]
ccgsnet: Solved at https://github.com/singnet/das/pull/1189/commits/33c997f5b72bfe6f83a622a5dac55ea385f0434b

src/tests/main/evaluation_evolution.cc
ccgsnet: Solved by 7898cf71d0b6b0e8e4425339ac36a1e7d6959c79

src/atomdb/inmemorydb/InMemoryDB.cc
ccgsnet: Solved by 7898cf71d0b6b0e8e4425339ac36a1e7d6959c79

src/atomdb/inmemorydb/InMemoryDB.cc
```diff
@@ -482,160 +511,184 @@ bool InMemoryDB::delete_link(const string& handle, bool delete_link_targets) {
...
+    lock_guard<mutex> lock(write_mutex_);
+    // Swap fresh tries instead of deleting in place: concurrent readers keep their
+    // pre-swap snapshots alive until they finish.
+    auto fresh_atoms = make_trie();
+    auto fresh_patterns = make_trie();
+    auto fresh_incoming = make_trie();
+    lock_guard<mutex> ptr_lock(trie_ptr_mutex_);
+    atoms_trie_ = fresh_atoms;
+    pattern_index_trie_ = fresh_patterns;
+    incoming_sets_trie_ = fresh_incoming;
```
andre-senna: It seems like you need to add a `{...}` block here otherwise `trie_ptr_mutex_` would be released at the end of the method together with `write_mutex_` but there's no way to assure the order in which they will be released. Adding the `{...}` block would assure `trie_ptr_mutex_` will be released first.

```suggestion
    {
        lock_guard<mutex> ptr_lock(trie_ptr_mutex_);
        atoms_trie_ = fresh_atoms;
        pattern_index_trie_ = fresh_patterns;
        incoming_sets_trie_ = fresh_incoming;
    }
```

andre-senna (approved):

src/atomdb/inmemorydb/InMemoryDB.cc
ccgsnet: Done https://github.com/singnet/das/pull/1189/commits/4a94d89125fadf15aa540bd18cf2ce35a836f81d

PR #1197, opened 2026-07-23
[#1196] Stream query answers as JSON over the Command Router HTTP API

### LINKED ISSUE #1196: Create a to_json() method in QueryAnswer.
use this method instead of `to_string(...)`

Image

marcocapozzoli:
Summary

- Adds `QueryAnswer::to_json()` to serialize QueryAnswers as structured JSON objects.
- Adds `QueryAnswer::from_json()` to rebuild a QueryAnswer from a JSON object.
- Updates the Command Router HTTP API WebSocket stream to send JSON answers instead of debug strings from to_string().
- Changes the chunk pipeline to use JSON arrays end-to-end (**BusCommandRouterProxyStreamPoller**, **CommandExecution**, **CommandRouterHttpAPI**).

`to_json()` with `metta_flag=false`.

```json
{
  "handles":[["289476dbf62faffc149e3658bc368b6b"]],
  "assignment":{"X":"a408f6dd446cdd4fa56f82e77fe6c870"},
  "metta_expressions":[],
  "assignment_metta":{},
  "importance":0.0,
  "strength":0.0
}
```

`to_json()`  with `metta_flag=true`.

```json
{
  "handles":[["289476dbf62faffc149e3658bc368b6b"]],
  "assignment": {"X":"a408f6dd446cdd4fa56f82e77fe6c870"},
  "metta_expressions": [["(Similarity \"human\" \"ent\")"]],
  "assignment_metta":{"X":"\"ent\""},
  "strength":0.0,
  "importance":0.0
}
```

Resolves #1196

### DIFF
--- a/src/agents/command_router/http_api/BusCommandRouterProxyStreamPoller.cc
+++ b/src/agents/command_router/http_api/BusCommandRouterProxyStreamPoller.cc
@@ -26,7 +26,10 @@ void append_answer_chunk(const shared_ptr<BusCommandRouterProxy>& router_proxy,
                          json& chunk_data) {
     shared_ptr<QueryAnswer> answer;
     while ((answer = router_proxy->pop()) != nullptr) {
-        bool use_metta_output = !answer->metta_expression.empty() || populate_metta_mapping;
+        if (populate_metta_mapping && answer->metta_expression.empty()) {
+            router_proxy->populate_metta_mapping(answer.get());
+        }
+        bool use_metta_output = populate_metta_mapping || !answer->metta_expression.empty();
         chunk_data.push_back(answer->to_json(use_metta_output));
         if (chunk_data.size() >= items_per_chunk) {
             emit_chunk(on_chunk, chunk_data);
--- a/src/agents/command_router/http_api/BusCommandRouterProxyStreamPoller.cc
+++ b/src/agents/command_router/http_api/BusCommandRouterProxyStreamPoller.cc
@@ -9,23 +9,25 @@ using namespace agents;
 
 namespace {
 
-void emit_chunk(const function<void(const vector<string>& chunk)>& on_chunk,
-                vector<string>& chunk_data) {
-    if (!chunk_data.empty() && on_chunk) {
-        on_chunk(chunk_data);
-        chunk_data.clear();
+void emit_chunk(const function<void(const json& chunk)>& on_chunk, json& chunk_data) {
+    if (chunk_data.empty()) {
+        return;
     }
+    if (on_chunk) {
+        on_chunk(chunk_data);
+    }
+    chunk_data = json::array();
 }
 
 void append_answer_chunk(const shared_ptr<BusCommandRouterProxy>& router_proxy,
                          bool populate_metta_mapping,
                          size_t items_per_chunk,
-                         const function<void(const vector<string>& chunk)>& on_chunk,
-                         vector<string>& chunk_data) {
+                         const function<void(const json& chunk)>& on_chunk,
+                         json& chunk_data) {
     shared_ptr<QueryAnswer> answer;
     while ((answer = router_proxy->pop()) != nullptr) {
         bool use_metta_output = !answer->metta_expression.empty() || populate_metta_mapping;
-        chunk_data.push_back(answer->to_string(use_metta_output));
+        chunk_data.push_back(answer->to_json(use_metta_output));
         if (chunk_data.size() >= items_per_chunk) {
             emit_chunk(on_chunk, chunk_data);
         }
@@ -44,7 +46,7 @@ PollStreamResult BusCommandRouterProxyStreamPoller::poll_stream(
     const string& command_type,
     size_t items_per_chunk,
     const function<bool()>& should_abort,
-    const function<void(const vector<string>& chunk)>& on_chunk,
+    const function<void(const json& chunk)>& on_chunk,
     const function<void(const string& error)>& on_error,
     const function<void()>& on_aborted) {
     if (items_per_chunk == 0) {
@@ -94,7 +96,7 @@ PollStreamResult BusCommandRouterProxyStreamPoller::poll_stream(
             return {};
         }
         if (on_chunk) {
-            on_chunk({router_proxy->params_response});
+            on_chunk(json::array({json(router_proxy->params_response)}));
         }
         return poll_succeeded();
     }
@@ -119,7 +121,7 @@ PollStreamResult BusCommandRouterProxyStreamPoller::poll_stream(
             return {};
         }
         if (on_chunk) {
-            on_chunk({router_proxy->set_param_ack});
+            on_chunk(json::array({json(router_proxy->set_param_ack)}));
         }
         return poll_succeeded();
     }
@@ -147,9 +149,9 @@ PollStreamResult BusCommandRouterProxyStreamPoller::poll_stream(
         const bool populate_metta_mapping =
             router_proxy->parameters.get_or<bool>(BaseQueryProxy::POPULATE_METTA_MAPPING, false);
 
-        vector<string> chunk_data;
+        json chunk_data = json::array();
         size_t streamed_item_count = 0;
-        auto emit_answer_chunk = [&](const vector<string>& chunk) {
+        auto emit_answer_chunk = [&](const json& chunk) {
             streamed_item_count += chunk.size();
             if (on_chunk) {
                 on_chunk(chunk);
--- a/src/agents/query_engine/QueryAnswer.cc
+++ b/src/agents/query_engine/QueryAnswer.cc
@@ -144,39 +144,135 @@ json QueryAnswer::to_json(bool metta_flag) {
     json answer = json::object();
 
     json handles_json = json::array();
-    string key = metta_flag ? "metta" : "handles";
+    json metta_handles_json = json::array();
     for (const vector<string>& group : this->handles) {
         json group_json = json::array();
+        json metta_group_json = json::array();
         for (const string& handle : group) {
+            group_json.push_back(handle);
             if (metta_flag) {
                 auto it = this->metta_expression.find(handle);
-                group_json.push_back(
+                metta_group_json.push_back(
                     (it != this->metta_expression.end() && !it->second.empty()) ? it->second : handle);
-            } else {
-                group_json.push_back(handle);
             }
         }
         handles_json.push_back(group_json);
-    }
-    answer[key] = handles_json;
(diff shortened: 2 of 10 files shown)


src/agents/command_router/http_api/BusCommandRouterProxyStreamPoller.cc
```diff
@@ -9,23 +9,25 @@ using namespace agents;
...
 void append_answer_chunk(const shared_ptr<BusCommandRouterProxy>& router_proxy,
                          bool populate_metta_mapping,
                          size_t items_per_chunk,
-                         const function<void(const vector<string>& chunk)>& on_chunk,
-                         vector<string>& chunk_data) {
+                         const function<void(const json& chunk)>& on_chunk,
+                         json& chunk_data) {
     shared_ptr<QueryAnswer> answer;
     while ((answer = router_proxy->pop()) != nullptr) {
         bool use_metta_output = !answer->metta_expression.empty() || populate_metta_mapping;
```
andre-senna: I understand that `populate_metta_mapping` is supposed to enforce that the answer need to be metta. So if `metta_expression` is empty and this flag is true, the JSON will be broken.

The correct behavior should be something like:

```
use_metta_output = false
if (populate_metta_mapping) {
    if (answer->metta_expression.empty()) {
        populates query answer with metta expressions
    }
    use_metta_output = true
}
```

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,45 @@ string QueryAnswer::to_string(bool metta_flag) {
     return answer;
 }
 
+json QueryAnswer::to_json(bool metta_flag) {
```
andre-senna, line 143: Add a test case for this method in QueryAnswer unit test using to_json() and from_json().

My suggestion is to change TEST(QueryAnswer, tokenization) to do both: tokenization/untokenization and to_json/from_json.

andre-senna, line 143: Add a from_json() method which takes a json and build a QueryAnswer out of it.

andre-senna (changes requested):

src/agents/query_engine/QueryAnswer.cc
andre-senna: from_json() won't be able to rebuild `metta_expression` even if the JSON is using metta expressions. 

In addition to this, if the JSON is based in metta, the rebuilding of a QueryAnswer won't be possible at all. So perhaps we want to build a JSON with both representations. That would be more adequate anyway because this way the JSON scheme would always be the same:

```
{
  "handles": [
    ["289476dbf62faffc149e3658bc368b6b"]
  ],
  "metta_expressions": [
    ["(Similarity \"human\" \"monkey\")"]
  ],
  "assignment": {
    "V": "a408f6dd446cdd4fa56f82e77fe6c870"
  },
 "assignment_metta": {
    "V": "\"monkey\""
  },
  "importance": 0.0,
  "strength": 0.0
}
```

When `to_json(metta_flag)` is called with `metta_flag==false`, "metta_expressions" and "assignment_metta" are kept empty. When `metta_flag==true`, everything is filled.

marcocapozzoli: Your JSON schema in `to_json(metta_flag)` treats the flag as an either/or output mode: with true, it clears handles/assignment and keeps only MeTTa. That works as a client view, but it can’t rebuild the same internal QueryAnswer, we lose the original handles and the handle -> metta map.

I thought of a JSON schema that always returns the full state (handles, assignment, metta_expression) and uses metta_flag only to fill convenience MeTTa views. Consumers pick what they need, and from_json can reconstruct the same object.

andre-senna: You misunderstood my last suggestion. I was suggesting to use this schema everywhere:

```
|{
  "handles": [
    ["289476dbf62faffc149e3658bc368b6b"]
  ],
  "metta_expressions": [
    ["(Similarity \"human\" \"monkey\")"]
  ],
  "assignment": {
    "V": "a408f6dd446cdd4fa56f82e77fe6c870"
  },
 "assignment_metta": {
    "V": "\"monkey\""
  },
  "importance": 0.0,
  "strength": 0.0
}
```

Exactly because of the problem you mentioned. As I said above, when to_json(metta_flag) is called with metta_flag==false, "metta_expressions" and "assignment_metta" are kept empty. When metta_flag==true, everything is filled.

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,143 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+    json metta_handles_json = json::array();
+    for (const vector<string>& group : this->handles) {
+        json group_json = json::array();
+        json metta_group_json = json::array();
+        for (const string& handle : group) {
+            group_json.push_back(handle);
+            if (metta_flag) {
+                auto it = this->metta_expression.find(handle);
+                metta_group_json.push_back(
+                    (it != this->metta_expression.end() && !it->second.empty()) ? it->second : handle);
```
andre-senna: Why adding `handle` if `it->second.empty()`? You should just check for emptiness before and add nothing if it's empty.

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,143 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+            }
+        }
+        handles_json.push_back(group_json);
+        if (metta_flag) {
```
andre-senna: `metta_group_json` is supposed to be added anyway. When metta_flag==false it will be added empty but it's still supposed to be added.

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,143 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+    }
+
+    json assignment_json = json::object();
+    json assignment_metta_json = json::object();
+    for (const auto& pair : this->assignment.table) {
+        assignment_json[pair.first] = pair.second;
+        if (metta_flag) {
+            auto it = this->metta_expression.find(pair.second);
+            assignment_metta_json[pair.first] =
+                (it != this->metta_expression.end() && !it->second.empty()) ? it->second : pair.second;
```
andre-senna: Again, you are supposed to check for emptiness before and add nothing if that's the case.

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,143 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+        }
+    }
+
+    answer["handles"] = handles_json;
+    answer["assignment"] = assignment_json;
+    answer["handle_to_metta"] = this->metta_expression;
+
+    if (metta_flag) {
```
andre-senna: This whole block is unnatural. I understand what you're doing but you're doing it in an unnatural way. You should just add the objects metta_handles_json and assignment_metta_json regardless if they are empty or not. Why re-allocating two new objects? Allocation is an expensive operation. Do you have any good reason to do it this way?

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,143 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+    if (!json_data.contains("importance") || !json_data["importance"].is_number()) {
+        RAISE_ERROR("Invalid QueryAnswer JSON: missing or invalid importance");
+    }
+    if (!json_data.contains("handles") || !json_data["handles"].is_array()) {
+        RAISE_ERROR("Invalid QueryAnswer JSON: missing or invalid handles");
+    }
+    if (!json_data.contains("assignment") || !json_data["assignment"].is_object()) {
+        RAISE_ERROR("Invalid QueryAnswer JSON: missing or invalid assignment");
+    }
+    if (!json_data.contains("handle_to_metta") || !json_data["handle_to_metta"].is_object()) {
```
andre-senna: Where did this handle_to_metta came from? We are not supposed to have handle_to_metta in the JSON.

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,143 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+                        "' must be a string");
+        }
+        if (!assignment.assign(it.key(), it.value().get<string>())) {
+            RAISE_ERROR("Invalid QueryAnswer JSON: conflicting assignment for '" + it.key() + "'");
+        }
+    }
+
+    const json& metta_json = json_data["handle_to_metta"];
+    map<string, string> metta_expression;
+    for (auto it = metta_json.begin(); it != metta_json.end(); ++it) {
```
andre-senna: We aren't supposed to add `QueryAnswer::metta_expression` to the JSON

src/tests/cpp/query_answer_test.cc
```diff
@@ -438,6 +438,125 @@ TEST(QueryAnswer, tokenization) {
     }
 }
 
+void query_answers_equal_including_metta(QueryAnswer* qa1, QueryAnswer* qa2) {
```
andre-senna, line 446: In addition to the test you added, please add a test case in the way I suggested in my previous comment.

andre-senna (changes requested):

src/agents/query_engine/QueryAnswer.cc
marcocapozzoli: You're right, that was my mistake

src/agents/query_engine/QueryAnswer.cc
```diff
@@ -140,6 +140,122 @@ string QueryAnswer::to_string(bool metta_flag) {
...
+        if (!it.value().is_string()) {
+            RAISE_ERROR("Invalid QueryAnswer JSON: assignment value for '" + it.key() +
+                        "' must be a string");
+        }
+        if (!assignment.assign(it.key(), it.value().get<string>())) {
+            RAISE_ERROR("Invalid QueryAnswer JSON: conflicting assignment for '" + it.key() + "'");
+        }
+    }
+
+    this->strength = strength;
```
andre-senna: I understand the point of having this "summarization" of everything that's required to build the QueryAnswer object but because we have this->handles and this->assignment (which are not scalar types) we are ending up doing unnecessary copies. Therefore, I suggest using "this->*" directly in each place above in order to avoid this unnecessary copy here.

andre-senna (approved):
@marcocapozzoli 
Please address my last comment before merging.

PR #1200, opened 2026-07-27
[#1195] Add parameters support to POST /command-router/executions

### LINKED ISSUE #1195: Update the "/command-router/executions" endpoint to accept parameters for Query.

marcocapozzoli:
This PR lets clients send router/query parameters in the HTTP execution request body, so you no longer need a separate set call before running a query.

**Changes**
- Added optional parameters object to POST /command-router/executions
- Parameters are applied via the existing router set param flow before the command runs
- Added build_set_param_arg() to validate parameter names and values against the command-router parameter schema
- Supports bool, integer, double, and string values (including JSON bool/number/string forms)
- Returns 400 for invalid parameter names, types, or values

**Example**
```json
{
  "command_type": "query",
  "command_text": "(Similarity \"human\" %C)",
  "parameters": {
    "populate_metta_mapping": true,
    "use_metta_as_query_tokens": "true",
  }
}
```

Resolves #1195

### DIFF
--- a/src/agents/command_router/http_api/CommandRouterHttpAPI.cc
+++ b/src/agents/command_router/http_api/CommandRouterHttpAPI.cc
@@ -160,27 +160,31 @@ void CommandRouterHttpAPI::setup_routes() {
 
                 LOG_INFO("CommandRouter HTTP API setting parameters for command=" << command_text);
 
+                Properties known_params =
+                    SystemParametersSingleton::get_instance()->get_command_router_params();
+                vector<pair<string, string>> set_args;
                 for (const auto& [key, value] : body["parameters"].items()) {
                     string validation_error;
-
                     const optional<string> args =
-                        this->build_set_param_arg(key, value, validation_error);
+                        this->build_set_param_arg(known_params, key, value, validation_error);
                     if (!args.has_value()) {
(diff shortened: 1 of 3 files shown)


ccgsnet (approved):

src/agents/command_router/http_api/CommandRouterHttpAPI.cc
```diff
@@ -150,6 +151,48 @@ void CommandRouterHttpAPI::setup_routes() {
...
+                        nullptr,
+                        [&](const string& message) { router_error = message; },
+                        nullptr);
+                    if (!poll_result.ok) {
+                        LOG_ERROR("CommandRouter HTTP API setting parameter failed for command="
+                                  << command_text << " key=" << key << " args=" << args
+                                  << " error=" << router_error);
+                        this->set_json_response(response, 500, {{"error", router_error}});
+                        return;
+                    }
```
andre-senna, line 191: This is wrong. Parameters being passed in a query request are supposed to be used ONLY in that query. No internal state for parameters values should be changed.

andre-senna (changes requested):

marcocapozzoli:
Outdated

Issue #1177, opened 2026-07-15
AtomDB Authorization


marcocapozzoli:

marcocapozzoli:
# Design proposal for data access control in AtomDB.

## Problem

DAS **does not authorize** access to data in AtomDB. Any client that knows the IP and port can connect directly and read everything, without going through the interface.

## Overall idea

Same pattern as `AdapterDB`: wrap any backend (`RedisMongoDB`, `MorkDB`, `RemoteAtomDB`, …) without touching their storage logic.

```
src/atomdb/auth/
    AuthorizationEntry.h / .cc          (authorization_types)
    AuthorizationManifest.h / .cc       (authorization_types)
    AuthorizationPersistence.h / .cc    (abstract + MongoAuthorizationPersistence)
    AuthorizationManagement.h / .cc
    ProtectedAtomDB.h / .cc             (exists today)
```

- MongoDB **generic config collection** (where index metadata is stored): boolean flag that the database is **protected**.
- MongoDB **`access_permissions`** (one per backend/peer): public key → LinkSchemas with read/write. The collection name is **not** configurable; it is a static in the backend that owns it.
- **`AuthorizationEntry`**: one `LinkSchema` plus independent read/write flags. Its `handle()` is the `LinkSchema` handle, which is what identifies an entry in `revoke(key, handle)`.
- **`AuthorizationManifest`**: in-RAM image of the whole `access_permissions` collection (`public_key` → one `Document` with `full_access` and its entries).
- **`AuthorizationPersistence`**: abstract storage interface for authorization writes. `MongoAuthorizationPersistence` is the concrete one and receives the Mongo pool, database name and collection name in its constructor. A future AtomDB type without Mongo only needs another concrete class here.
- **`AuthorizationManagement`**: the authorization query/administration API. Keeps an `AuthorizationManifest` in RAM, builds it in the constructor, and keeps Mongo and RAM in sync on every `authorize()` / `revoke*()`.
- **`ProtectedAtomDB`**: AtomDB wrapper installed when the backend is protected. Data-access methods have two forms: without `public_key` (rejected) and with `public_key` (authorize, then delegate to the backend). Never talks to Mongo for permissions itself.
- **`AtomDBFactory`**: creates the concrete backend, asks `is_protected()`, and wraps it with `ProtectedAtomDB` when needed. It is also the single place that knows the concrete backend type, so it builds the matching `AuthorizationPersistence`.
[...]

PR #1224, opened 2026-08-12
[#1177] Add AtomDB::get_access_permissions()

marcocapozzoli:
Adds AtomDB::get_access_permissions() to read access permission docs from the DB.

RedisMongoDB loads them from Mongo. The other AtomDBs either forward the call or return empty.


src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -32,6 +32,7 @@ string RedisMongoDB::MONGODB_DB_NAME;
 string RedisMongoDB::MONGODB_NODES_COLLECTION_NAME;
 string RedisMongoDB::MONGODB_LINKS_COLLECTION_NAME;
 string RedisMongoDB::MONGODB_PATTERN_INDEX_SCHEMA_COLLECTION_NAME;
+string RedisMongoDB::MONGODB_ACCESS_PERMISSIONS_COLLECTION_NAME;
```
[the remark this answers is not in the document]
marcocapozzoli: I'm following the existing RedisMongoDB pattern, where collection names are defined via `initialize_statics(context)`. Making only `access_permissions` an instance attribute would be inconsistent; if we were to make that change, it should be part of a broader refactoring of all these static elements, which is outside the scope of this PR.

src/atomdb/redis_mongodb/RedisMongoDB.h
```diff
@@ -41,6 +42,7 @@ class RedisMongoDB : public AtomDB {
     static string MONGODB_NODES_COLLECTION_NAME;
     static string MONGODB_LINKS_COLLECTION_NAME;
     static string MONGODB_PATTERN_INDEX_SCHEMA_COLLECTION_NAME;
+    static string MONGODB_ACCESS_PERMISSIONS_COLLECTION_NAME;
```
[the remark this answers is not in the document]
marcocapozzoli: I'm following the existing RedisMongoDB pattern, where collection names are defined via `initialize_statics(context)`. Making only `access_permissions` an instance attribute would be inconsistent; if we were to make that change, it should be part of a broader refactoring of all these static elements, which is outside the scope of this PR.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -39,6 +39,18 @@ bool RemoteAtomDBPeer::allow_nested_indexing() {
...
 }
+vector<string> RemoteAtomDBPeer::get_access_permissions() const {
+    vector<string> documents;
+    if (local_persistence_) {
+        auto local_documents = local_persistence_->get_access_permissions();
+        documents.insert(documents.end(), local_documents.begin(), local_documents.end());
+    }
+    if (atomdb_) {
+        auto remote_documents = atomdb_->get_access_permissions();
+        documents.insert(documents.end(), remote_documents.begin(), remote_documents.end());
```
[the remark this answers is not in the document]
marcocapozzoli: Not worth it here, the list is tiny, so this change wouldn’t make a real difference. We can leave it as is.

src/atomdb/AtomDB.h
```diff
@@ -112,6 +113,14 @@ class AtomDB : public HandleDecoder {
...
 
     bool empty() const { return atom_count() == 0; }
+
+   protected:
+    // AdapterDB / RemoteAtomDBPeer call this on a composed AtomDB*, not on this.
+    friend class AdapterDB;
+    friend class RemoteAtomDBPeer;
+
+    virtual vector<shared_ptr<atomdb_api_types::AccessPermissionDocument>> get_access_permissions()
+        const = 0;
```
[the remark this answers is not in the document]
marcocapozzoli: outdated

src/atomdb/remotedb/RemoteAtomDB.h
```diff
@@ -80,6 +80,11 @@ class RemoteAtomDB : public AtomDB {
 
     void release_caches(const LinkSchema& link_schema, bool persist = true, bool force = false);
 
+    vector<atomdb_api_types::AccessPermissionDocument> get_access_permissions(
+        const atomdb_api_types::PublicKey& public_key) const override {
+        return {};
+    };
```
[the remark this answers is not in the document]
marcocapozzoli: Intentional. RemoteAtomDB does not look up permissions; it always returns empty. The federation test asserts that.

src/atomdb/AtomDB.h
```diff
@@ -112,6 +113,9 @@ class AtomDB : public HandleDecoder {
     virtual size_t atom_count() const = 0;
 
     bool empty() const { return atom_count() == 0; }
+
+    virtual vector<atomdb_api_types::AccessPermissionDocument> get_access_permissions(
+        const atomdb_api_types::PublicKey& public_key) const = 0;
```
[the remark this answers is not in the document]
marcocapozzoli: solve in 04ed27c39eab1551fd084b12e7e81c2cf3c21d01

src/atomdb/remotedb/RemoteAtomDBPeer.h
```diff
@@ -30,6 +30,8 @@ class RemoteAtomDBPeer : public AtomDB, public processor::ThreadMethod {
 
     bool allow_nested_indexing() override;
     bool composite_type_enabled() const override;
+    vector<shared_ptr<atomdb_api_types::AccessPermissionDocument>> get_access_permissions()
+        const override;
```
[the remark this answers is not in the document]
marcocapozzoli: outdated

src/atomdb/AtomDB.h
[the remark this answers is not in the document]
marcocapozzoli: it's not necessary

src/atomdb/AtomDBAPITypes.h
```diff
@@ -67,5 +74,38 @@ class AtomDocument {
...
+class AccessPermissionDocument {
+   public:
+    string public_key;
+    bool full_access;
+    vector<AccessPermissionEntry> entries;
+
+    AccessPermissionDocument(const string& public_key,
+                             bool full_access,
+                             const vector<AccessPermissionEntry>& entries)
+        : public_key(public_key), full_access(full_access), entries(entries) {}
```
andre-senna: Add a check to make sure that:

* (1) If `full_access` is true, `entries` must be empty
* (2) If `full_access` is false, `entries` must be non-empty

src/atomdb/AtomDBAPITypes.h
```diff
@@ -67,5 +74,38 @@ class AtomDocument {
...
+    AccessPermissionEntry(const LinkSchema& schema, bool read, bool write)
+        : schema(schema), read(read), write(write) {}
+
+    AccessPermissionEntry(const vector<string>& tokens, bool read, bool write)
+        : schema(tokens), read(read), write(write) {}
+};
+
+class AccessPermissionDocument {
+   public:
+    string public_key;
```
andre-senna: There's a PublicKey class below. So I think you meant to use it here instead of `string`? If you really want to use `string` here you should change the name of this variable.

src/atomdb/AtomDBAPITypes.h
```diff
@@ -67,5 +74,38 @@ class AtomDocument {
...
+
+    AccessPermissionDocument(const string& public_key,
+                             bool full_access,
+                             const vector<AccessPermissionEntry>& entries)
+        : public_key(public_key), full_access(full_access), entries(entries) {}
+};
+
+class PublicKey {
+   public:
+    variant<string, map<string, string>> value;
```
andre-senna: `variant` types, although a powerful syntax sugar, are awkward and error prone. Their use are more natural in some contexts like multi-typed dicts or designs where the return type of chained functions are supposed to change. When the `variant` is used in parameters instead of return-values, other design choices should be made (like e.g. plain inheritance).

In this particular case I may suggest an alternative simple design like these: store actual string keys in a vector<string> and use a map<string, unsigned int> to map from peer names (keys) to positions in this array. This way a PublicKey with map.size() == 0 is a simple key while one with map.size() > 0 is a key used in a RemoteDB. A simple key is stored in the vector<string> in the same way as the keys from RemoteDB.

marcocapozzoli:
@coderabbitai full review

src/atomdb/inmemorydb/InMemoryDB.h
```diff
@@ -41,6 +41,14 @@ class InMemoryDB : public AtomDB {
 
     bool allow_nested_indexing() override;
     bool composite_type_enabled() const override { return false; }
+    /**
+     * Access permissions are not stored in this AtomDB.
+     * Always returns an empty vector.
+     */
+    vector<atomdb_api_types::AccessPermissionDocument> get_access_permissions(
```
andre-senna: We should implement this empty `get_access_permission()` in the abstract AtomDB class. Remember to make it `virtual` so any calls will prioritize forwarding to subclasses. This way we can remove it from all subclasses that don't actually need it.

src/atomdb/remotedb/RemoteAtomDBPeer.cc
```diff
@@ -39,6 +39,11 @@ bool RemoteAtomDBPeer::allow_nested_indexing() { return atomdb_->allow_nested_in
 bool RemoteAtomDBPeer::composite_type_enabled() const {
     return local_persistence_ && local_persistence_->composite_type_enabled();
 }
+vector<atomdb_api_types::AccessPermissionDocument> RemoteAtomDBPeer::get_access_permissions(
+    const atomdb_api_types::PublicKey& public_key) const {
+    if (!atomdb_) return {};
```
andre-senna: Considering the comment I made for `InMemoryDB` (regarding putting an empty implementation in AtomDB), we should change this to:


```suggestion
    if (!atomdb_) return AtomDB::get_access_permissions(public_key);
```

andre-senna (changes requested):

src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -55,6 +59,88 @@ RedisMongoDB::~RedisMongoDB() {
...
+            if (!token.is_string() || token.get<string>().empty()) {
+                RAISE_ERROR("AccessPermissionDocument allowed_schemas tokens must be non-empty strings");
+            }
+            tokens.push_back(token.get<string>());
+        }
+
+        entries.emplace_back(tokens, item["read"].get<bool>(), item["write"].get<bool>());
+    }
+
+    return atomdb_api_types::AccessPermissionDocument(key, document["full_access"].get<bool>(), entries);
```
ccgsnet: I think we should use `document["public_key"]` instead of `key` here. Even knowing that (for now) both have the same value.

src/atomdb/redis_mongodb/BUILD
```diff
@@ -42,6 +42,7 @@ cc_library(
     deps = [
         "//atomdb:atomdb_api_types",
         "//commons/atoms:atoms_lib",
+        "@nlohmann_json//:json",
```
ccgsnet: Move this to the `redis_mongodb` deps

src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -55,6 +59,88 @@ RedisMongoDB::~RedisMongoDB() {
...
+        if (document.has_value()) {
+            documents.push_back(document.value());
+        }
+    }
+    return documents;
+}
+
+optional<atomdb_api_types::AccessPermissionDocument> RedisMongoDB::load_access_permission_document(
+    const string& key) const {
+    string handle = compute_hash((char*) key.c_str());
```
ccgsnet: Avoid leak: `string handle = Hasher::plain_string_hash(key);`

src/atomdb/AtomDBAPITypes.h
```diff
@@ -1,14 +1,21 @@
 #pragma once
 
 #include <algorithm>
+#include <map>
+#include <memory>
 #include <optional>
+#include <string>
+#include <variant>
```
ccgsnet: It seems we can remove these includes:
```
#include <memory>
#include <optional>
#include <variant>
```

ccgsnet (approved):

src/atomdb/AtomDBAPITypes.h
andre-senna: You got it wrong. I'm not suggesting any side-effect on the inputs, I'm just asking you to make checks to assure the coherence of the inputs.  Something like:

```
if ((full_access && (entries.size() > 0)) || (!full_access && (entries.size() == 0)) { 
    RAISE_ERROR(...)
}
```

andre-senna (changes requested):

src/atomdb/AtomDBAPITypes.h
marcocapozzoli: I got it now.

andre-senna (approved):

PR #1239, opened 2026-08-25
Refactor access permission types to abstract API

marcocapozzoli:
Required by #922 
## Summary

- Turn `AccessPermissionEntry` and `AccessPermissionDocument` into abstract interfaces with virtual getters instead of public fields
- Add read-only MongoDB implementations: `MongodbAccessPermissionEntry` and `MongodbAccessPermissionDocument`
- Update `RedisMongoDB` to load permissions directly from BSON (removes JSON parsing path)
- Change `get_access_permissions()` to return `vector<shared_ptr<AccessPermissionDocument>>`

### DIFF
--- a/src/atomdb/redis_mongodb/RedisMongoDBAPITypes.cc
+++ b/src/atomdb/redis_mongodb/RedisMongoDBAPITypes.cc
@@ -269,3 +269,96 @@ MongodbDocument::MongodbDocument(const atoms::Link* link,
     add_custom_attributes(doc, link->custom_attributes);
     this->document = doc.extract();
 }
+
+MongodbAccessPermissionEntry::MongodbAccessPermissionEntry(bsoncxx::document::view entry_view)
+    : entry_view_(entry_view) {}
+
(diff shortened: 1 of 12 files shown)


src/atomdb/redis_mongodb/RedisMongoDBAPITypes.cc
```diff
@@ -269,3 +269,96 @@ MongodbDocument::MongodbDocument(const atoms::Link* link,
...
+
+bool MongodbAccessPermissionEntry::get_read() const {
+    return this->entry_view_["read"].get_bool().value;
+}
+
+bool MongodbAccessPermissionEntry::get_write() const {
+    return this->entry_view_["write"].get_bool().value;
+}
+
+unsigned int MongodbAccessPermissionEntry::get_tokens_size() const {
```
andre-senna, line 284: Please add a test case specifically for this method getting the size of token array with different sizes including, at least, size = {0, 1, 2}.

andre-senna (approved):

PR #1249, opened 2026-09-01
[#1208] Miscellaneous new features and fixes in LCA and related code

### LINKED ISSUE #1208: Review LCA implementation to adapt to the new algorithm in the evaluation_evolution POC
See algorithms in build_links() and downstream method calls in evaluation_evolution_main.cc

See original algorithm here: https://github.com/singnet/das/issues/128?issue=singnet%7Cdas%7C209

andre-senna:
Resolves #1208 

This PR concludes the work in #1208. Here we have the final features and fixes required to have a basic version of the LCA running with only on built-in link creation function. In addition to the changes in the LCA package, we made a couple of changes in some other packages as well as we fixed the issues found in tests. This is why this PR is so long.

We also add an integration test which covers most of LCA functionalities.

We'll implement the other required link creation functions in a follow-up PR related to #1244


src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
```diff
@@ -0,0 +1,104 @@
...
+    if (predicates[0] != predicates[1]) {
+        if (!visited(key)) {
+            visit(key);
+            stats.visited = true;
+            set<string> mentioned_predicates0, mentioned_predicates1;
+            extract_mentioned_predicates(mentioned_predicates0, predicates[0]);
+            extract_mentioned_predicates(mentioned_predicates1, predicates[1]);
+            if (!Utils::intersects(mentioned_predicates0, mentioned_predicates1)) {
+                vector<string> targets = {LOGICAL_AND_HANDLE, predicates[0], predicates[1]};
+                if (add_or_update_link(targets, 1.0)) {
```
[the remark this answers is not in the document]
andre-senna: Correct but this is the expected behavior.

src/agents/link_creation_agent/LinkCreationProxy.cc
```diff
@@ -24,12 +29,14 @@ LinkCreationProxy::LinkCreationProxy() {
 LinkCreationProxy::LinkCreationProxy(const vector<string>& tokens,
                                      const string& context,
                                      const string& link_creator_tag,
+                                     BaseProxy::ORCHESTRATION_SCHEMA_TYPE orchestration,
                                      const shared_ptr<LinkCreator> link_creation_function)
     : BaseQueryProxy(tokens, context) {
     // constructor typically used in requestor
     init();
     this->link_creation_function_object = link_creation_function;
     set_link_creator_function_tag(link_creator_tag);
```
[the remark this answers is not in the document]
andre-senna: That constructor is used only in client side and when a link_creation_function object is passed, it's supposed to be a remote function so the tag is required to be LinkCreatorRegistry::REMOTE_FUNCTION. The proper checking is already performed inside set_link_creator_function_tag(). As for a test case for this, we require remote functions implemented before making a gook test case so this test case will be postponed until then.

src/tests/integration/cpp/lca_integration_test.cc
```diff
@@ -0,0 +1,215 @@
+#include "AndTwoPredicates.h"
+#include "AtomDBSingleton.h"
+#include "JsonConfigParser.h"
+#include "LinkCreationProxy.h"
+#include "LinkCreatorRegistry.h"
+#include "Logger.h"
```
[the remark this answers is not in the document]
andre-senna: No. We are supposed to use the default LOG_LEVEL defined in Logger.h

src/tests/scripts/lca_test_environment.sh
```diff
@@ -0,0 +1,45 @@
...
+das-cli db stop --prune >> /tmp/atomdb.log 2>&1
+(docker stop `docker ps | tail -n +2 | cut -d" " -f1` ; docker container prune -f ; docker ps -a) > /dev/null 2>&1
+das-cli db start >> /tmp/atomdb.log 2>&1
+make run-db-loader OPTIONS="--config=/opt/das/config/das.json --file=/tmp/$KB --threads=4 --chunk=5000" >> /tmp/atomdb.log 2>&1
+make run-attention-broker >> /tmp/attention_broker.log 2>&1 &
+sleep 5
+make run-busnode OPTIONS="--service=query-engine --config=/opt/das/config/das.json" >> /tmp/query_engine.log 2>&1 &
+sleep 5
+make run-busnode OPTIONS="--service=link-creation-agent --bus-endpoint=localhost:40002 --config=/opt/das/config/das.json" >> /tmp/link_creation_agent.log 2>&1 &
+sleep 5
```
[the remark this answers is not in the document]
andre-senna: Although the finding is valid and correct, fixing it would be overkill. If any of the services takes longer than 5 seconds to get up, it's almost surely in error, so it's OK if the test fails in this situation (which will happen).

src/commons/SystemParametersValidation.cc
```diff
@@ -35,19 +35,15 @@ const AgentParamsSchema& params_schema() {
...
-         {{"inference_request_timeout", "long"},
-          {"repeat_count", "unsigned_int"},
-          {"max_answers", "unsigned_int"}}},
+          {"disregard_importance_flag", "bool"},
+          {"unique_value_flag", "bool"},
+          {"max_successful_creation_per_round", "unsigned_int"},
+          {"max_unproductive_visits_per_round", "unsigned_int"},
+          {"max_visit_attempts_per_round", "unsigned_int"},
+          {"max_rounds", "unsigned_int"},
+          {"link_creation_strength_threshold", "double"}}},
```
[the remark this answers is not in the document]
andre-senna: This is the expected behavior. link_creation_log_file_name is not supposed to be in the JSON config file but it can be set in runtime in the proxy before issuing a command.

src/agents/link_creation_agent/link_creators/tags.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+// Query tokens
+static string AND_OPERATOR = string("AND");
```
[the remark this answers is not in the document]
andre-senna: Fixed in d3887ef67f9109d0e477fb7f5920341ca3c10341

src/agents/link_creation_agent/LinkCreationProcessor.cc
```diff
@@ -111,15 +134,35 @@ void LinkCreationProcessor::remove_processor_thread(const string& stoppable_thre
...
+
 void LinkCreationProcessor::link_creation(shared_ptr<StoppableThread> monitor,
                                           shared_ptr<LinkCreationProxy> proxy) {
     STACK_TRACE();
     unsigned int count_created = 0;
+    unsigned int count_used_query_answers = 0;
+    unsigned int count_query_answers_in_cycle = 0;
+    unsigned int count_iterated_query_answers = 0;
     unsigned int unproductive_visit = 0;
     unsigned int visit_attempts = 0;
```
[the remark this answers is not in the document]
andre-senna: Fixed in d3887ef67f9109d0e477fb7f5920341ca3c10341

src/tests/scripts/lca_test_environment.sh
```diff
@@ -0,0 +1,47 @@
...
+(docker stop `docker ps | tail -n +2 | cut -d" " -f1` ; docker container prune -f ; docker ps -a) > /dev/null 2>&1
+das-cli db start >> ${TMP_DIR}/atomdb.log 2>&1
+make run-db-loader OPTIONS="--config=/opt/das/config/das.json --file=${TMP_DIR}/$KB --threads=4 --chunk=5000" >> ${TMP_DIR}/atomdb.log 2>&1
+make run-attention-broker >> ${TMP_DIR}/attention_broker.log 2>&1 &
+sleep 5
+make run-busnode OPTIONS="--service=query-engine --config=/opt/das/config/das.json" >> ${TMP_DIR}/query_engine.log 2>&1 &
+sleep 5
+make run-busnode OPTIONS="--service=link-creation-agent --bus-endpoint=localhost:40002 --config=/opt/das/config/das.json" >> ${TMP_DIR}/link_creation_agent.log 2>&1 &
+sleep 5
+\rm -rf ${TMP_DIR}
```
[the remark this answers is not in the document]
andre-senna: Fixed in cdf83710be19f5fb6c236a2ebb4babfbf3de6510

andre-senna:
@coderabbit review

andre-senna:
@coderabbitai approve

src/agents/link_creation_agent/link_creators/LinkCreator.cc
```diff
@@ -0,0 +1,80 @@
...
+        return "";
+    } else {
+        return node->name;
+    }
+}
+
+double LinkCreator::get_strength(const string& handle) {
+    STACK_TRACE();
+    auto atom = atomdb()->get_atom(handle);
+    return atom->custom_attributes.get_or<double>(STRENGTH_TAG, 1.0);
```
ccgsnet: `nullptr` check before returning

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
```diff
@@ -0,0 +1,105 @@
...
+    string concept_ = query_answer->get(CONCEPT);
+    string predicates[2];
+    predicates[0] = query_answer->get(PREDICATE1);
+    predicates[1] = query_answer->get(PREDICATE2);
+    if (predicates[1] < predicates[0]) {
+        string aux = predicates[0];
+        predicates[0] = predicates[1];
+        predicates[1] = aux;
+    }
+    string key = predicates[0] + " " + predicates[1];
```
ccgsnet: Don't we need to add `concept` to the `key`?
If we get a `P1 P2 C1` and then a `P1 P2 C2`, wouldn't we miss the second one?

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
```diff
@@ -0,0 +1,105 @@
...
+    if (predicates[0] != predicates[1]) {
+        if (!visited(key)) {
+            visit(key);
+            stats.visited = true;
+            set<string> mentioned_predicates0, mentioned_predicates1;
+            extract_mentioned_predicates(mentioned_predicates0, predicates[0]);
+            extract_mentioned_predicates(mentioned_predicates1, predicates[1]);
+            if (!Utils::intersects(mentioned_predicates0, mentioned_predicates1)) {
+                vector<string> targets = {LOGICAL_AND_HANDLE, predicates[0], predicates[1]};
+                if (add_or_update_link(targets, 1.0)) {
```
ccgsnet: Same thing here, if we already inserted `(LogicalAnd P1 P2)` in a previous round (eg for `C1`), we would miss a new `Evaluation` link for a different `C2`, wouldn't we?

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
```diff
@@ -0,0 +1,105 @@
...
+                              vector<string>(mentioned_predicates0.begin(), mentioned_predicates0.end()),
+                              '-') +
+                          ", " +
+                          Utils::join(
+                              vector<string>(mentioned_predicates1.begin(), mentioned_predicates1.end()),
+                              '-') +
+                          ") " + "Skipping link building because predicates intersect.");
+            }
+        } else {
+            LOG_DEBUG(
```
ccgsnet: Don't we need to ensure `stats.visited = true;` before returning here too?

ccgsnet (approved):

src/agents/link_creation_agent/link_creators/LinkCreator.cc
andre-senna: Fixed

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
andre-senna: Fixed

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
andre-senna: Actually no. This check is a quite different one. It only prevents us from reusing a predicate we already added to an AND clause. For instance if we already have predicates PA = (p1 AND p2),  PB = (p3 AND p4) and PC = (p1 AND p5), we want to visit (PA AND PB) as a potential candidate to build the link but we don't want to visit (PA AND PC) because they share the same inner predicate p1. The check is written to make this recursively so any nested predicates would work the same way.

So no fix is required.

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
andre-senna: No. "visited" means that a potentially expensive operation has been performed in order to attempt a link creation. The point of having the "visited" flag is to limit the number of such potentially expensive operations. So if we skipped the candidate, this counter is supposed to be increased.

No fix required.

src/agents/link_creation_agent/link_creators/AndTwoPredicates.cc
andre-senna: On a second thought, you are correct. I was talking about the check in line 38 but what you really meant is the check in line 40. Fixed in 475aace416e414b09405fe6cc04845c8c27b0d65

PR #1259, opened 2026-09-08
Remove `AtomDB` context and use prefix for `RedisMongoDB`

ccgsnet:
## Summary

- Remove the AtomDB `context` constructor/factory argument. Redis/Mongo namespace isolation is now optional `JsonConfig["prefix"]` on each `redismongodb` / `morkdb` backend (including remotedb peers and `adapterdb.atomdb_backend`).
- Redis key prefixes and MongoDB database/collection names are per `RedisMongoDB` instance, so two backends can use different prefixes in the same process.
- Mutating tests pass a unique prefix so they no longer share or wipe the default `das` database used by query tests.

## Breaking change

`AtomDBFactory::create()` (and the Redis/Mork/InMemory/Adapter constructors) no longer take `context`. Callers that need isolation should set `"prefix"` on the backend config, e.g. `"test_"` → Mongo DB `test_das` and Redis keys `test_patterns`, `test_outgoing_set`, `test_incoming_set`. Omit it for the default names.

## Test plan

- [ ] `make test-all` (or at least `//tests/cpp:nested_link_template_test`, `redis_mongodb_test`, `remote_atomdb_test`, `adapterdb_test`, `morkdb_test`)
- [ ] Two RedisMongoDB instances with different prefixes keep independent namespaces
- [ ] Remotedb with two RedisMongoDB peers and different prefixes stays isolated

### DIFF
--- a/src/atomdb/redis_mongodb/RedisMongoDB.cc
+++ b/src/atomdb/redis_mongodb/RedisMongoDB.cc
@@ -40,13 +40,13 @@ string RedisMongoDB::MONGODB_ACCESS_PERMISSIONS_COLLECTION_NAME;
 string RedisMongoDB::MONGODB_FIELD_NAME[MONGODB_FIELD::size];
 uint RedisMongoDB::MONGODB_CHUNK_SIZE;
 
-RedisMongoDB::RedisMongoDB(const string& context, bool skip_redis, const JsonConfig& config)
-    : context(context),
-      skip_redis_(skip_redis),
+RedisMongoDB::RedisMongoDB(const JsonConfig& config)
+    : skip_redis_(config.at_path("type").get_or<string>("") != "redismongodb"),
       composite_type_enabled_(config.at_path("composite_type_enabled").get_or<bool>(true)),
       cluster_flag(false),
       protection_mode(atomdb_api_types::ProtectionMode::PROTECTED) {
-    initialize_statics(context);
+    string prefix = config.at_path("prefix").get_or<string>("");
+    initialize_statics(prefix);
     mongodb_setup(config);
     load_protection_mode();
     load_pattern_index_schema();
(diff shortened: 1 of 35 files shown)


src/atomdb/redis_mongodb/RedisMongoDB.cc
```diff
@@ -27,26 +27,35 @@ using namespace atomdb;
...
-string RedisMongoDB::MONGODB_CONFIG_COLLECTION_NAME;
-string RedisMongoDB::MONGODB_PATTERN_INDEX_SCHEMA_COLLECTION_NAME;
-string RedisMongoDB::MONGODB_ACCESS_PERMISSIONS_COLLECTION_NAME;
 string RedisMongoDB::MONGODB_FIELD_NAME[MONGODB_FIELD::size];
 uint RedisMongoDB::MONGODB_CHUNK_SIZE;
 
-RedisMongoDB::RedisMongoDB(const string& context, bool skip_redis, const JsonConfig& config)
-    : context(context),
-      skip_redis_(skip_redis),
+void RedisMongoDB::initialize_namespace(const string& prefix) {
```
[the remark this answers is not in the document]
ccgsnet: Added at [733bef0](https://github.com/singnet/das/pull/1259/commits/733bef0c60dabd5298c24c4af52279aa02ce0fcf)

src/scripts/test_all_setup.sh
```diff
@@ -15,7 +15,7 @@ DAS_MORK_HOSTNAME=0.0.0.0
 DAS_MORK_PORT=40022
 
 REDIS_IMAGE=redis:7.2.3-alpine
-MONGODB_IMAGE=mongodb/mongodb-community-server:8.2-ubuntu2204
+MONGODB_IMAGE=mongodb/mongodb-community-server:8.0.4-ubuntu2204
```
andre-senna: Is this intentional? I mean, keeping the change in the PR?

andre-senna (approved):

src/scripts/test_all_setup.sh
ccgsnet: No, reverting that now...

PR #1271, opened 2026-09-10
Add uid to `AtomDB`

ccgsnet:
## Summary

- Require `uid` on AtomDB config (`atomdb.uid` in `das.json`, factory `create()`, remotedb peers). The value may be empty.
- `get_uid()` exposes it. Internal `InMemoryDB()` caches also use an empty uid.
- `PublicKey::key_for_uid()` selects this instance’s key; `RedisMongoDB::get_access_permissions()` uses that instead of scanning every peer key.

## Test plan

- [ ] `//tests/cpp:inmemorydb_test`, `protected_atomdb_test`, `config_parser_test`, `redis_mongodb_test`
- [ ] `//tests/cpp:atomdb_factory_test`, `remote_atomdb_test` (need MORK)
- [ ] Startup fails if `atomdb.uid` is missing; `"uid": ""` is accepted

### DIFF
--- a/src/atomdb/AtomDBFactory.cc
+++ b/src/atomdb/AtomDBFactory.cc
@@ -15,6 +15,10 @@ using namespace commons;
 // Public methods
 
 shared_ptr<AtomDB> AtomDBFactory::create(const JsonConfig& config) {
+    if (config.at_path("uid").is_null()) {
+        RAISE_ERROR("AtomDBFactory: missing uid");
+    }
+
     auto atomdb_type = config.at_path("type").get_or<string>("");
 
     AtomDBType type = AtomDB::string_to_type(atomdb_type);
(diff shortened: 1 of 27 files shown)


src/scripts/test_all_setup.sh
```diff
@@ -15,7 +15,7 @@ DAS_MORK_HOSTNAME=0.0.0.0
 DAS_MORK_PORT=40022
 
 REDIS_IMAGE=redis:7.2.3-alpine
-MONGODB_IMAGE=mongodb/mongodb-community-server:8.2-ubuntu2204
+MONGODB_IMAGE=mongodb/mongodb-community-server:8.0.4-ubuntu2204
```
andre-senna: Leftover?

andre-senna (approved):

src/atomdb/AtomDBFactory.cc
```diff
@@ -37,6 +41,10 @@ shared_ptr<AtomDB> AtomDBFactory::create(const JsonConfig& config) {
 // Private methods
 
 shared_ptr<AtomDB> AtomDBFactory::create_basic_atomdb(const JsonConfig& config) {
+    if (config.at_path("uid").is_null()) {
```
marcocapozzoli: I don't think this check is necessary here. It's already performed in the public `::create()` method.

marcocapozzoli (approved):

src/scripts/test_all_setup.sh
ccgsnet: Reverted at [04c40e6](https://github.com/singnet/das/pull/1271/commits/04c40e69cc708cdf941fadf23a98b98631e953ad)

PR #1276, opened 2026-09-14
[#1244] Add CustomizableLinkCreator to LCA

andre-senna:
WIP towards #1244 

In this PR we add a new type of Link Creator to LCA. CustomizableLinkCreator uses a spec passed through LINK_CREATOR_EXTRA_PARAMETERS to c create links using the elements of the QueryAnswer objects returned by the Query Agent.

Additionally we add a new minor feature to Utils::join() to allow the use of string as separator (this new feature was used in the tests). Minor fixes to previously added AndTwoPredicates were also made.


src/agents/link_creation_agent/link_creators/CustomizableLinkCreator.cc
```diff
@@ -0,0 +1,136 @@
...
+                stats.updated++;
+            }
+        }
+    }
+    return stats;
+}
+
+void CustomizableLinkCreator::extra_parameters(const string& extra_parameters) {
+    if (extra_parameters != "") {
+        vector<string> tokens = Utils::split(extra_parameters);
```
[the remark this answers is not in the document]
andre-senna: @coderabbitai Fixed in [88987df](https://github.com/singnet/das/pull/1276/commits/88987df8657edba1babc350e76cee9754e8ecc86)

src/agents/link_creation_agent/link_creators/CustomizableLinkCreator.cc
```diff
@@ -0,0 +1,136 @@
...
+CustomizableLinkCreator::CustomizableLinkCreator() {}
+
+CustomizableLinkCreator::~CustomizableLinkCreator() {}
+
+LinkCreationStats CustomizableLinkCreator::create(shared_ptr<QueryAnswer> query_answer) {
+    STACK_TRACE();
+    LinkCreationStats stats;
+    for (LinkSpecification& spec : this->link_specification) {
+        if ((spec.target_elements.size() == 0) || (spec.link_type == "")) {
+            RAISE_ERROR("Invalid empty target elements or link_type");
```
ccgsnet, line 20: Do we really want to return `stats` here? If so we should use `LOG_ERROR` and not `RAISE`

src/agents/link_creation_agent/link_creators/CustomizableLinkCreator.cc
```diff
@@ -0,0 +1,136 @@
...
+            handles.push_back(query_answer->get(element));
+        }
+        string key = Utils::join(handles, ' ');
+        if (!visited(key)) {
+            visit(key);
+            stats.visited = true;
+            for (QueryAnswerElement& element : spec.strength_elements) {
+                strength_components.push_back(get_strength(query_answer->get(element)));
+            }
+            if (add_or_update_link(handles,
```
ccgsnet: Dont we need to check `if (strength >= strength_threshold()) {` before `add_or_update_link()` here?

ccgsnet (approved):

src/agents/link_creation_agent/link_creators/CustomizableLinkCreator.cc
andre-senna: Even when raising, I like to keep the code returning the best way it can. But I agree it's bit weird in this context so I just replaced the return by a break. It sounds less weird and the return of the empty stats will happen anyway in the end of the function.

src/agents/link_creation_agent/link_creators/CustomizableLinkCreator.cc
andre-senna: Fixed.

You are correct. To prevent the same mistake from happening again when new link creators are added, I moved the check to inside the add_or_update_link() method, which is where it should be in the first place.

PR #1280, opened 2026-09-17
[#1277] Implement ProtectedAtomDB read methods

### LINKED ISSUE #1277: Implement ProtectedAtomDB.cc

marcocapozzoli:
- Implement the read path in `ProtectedAtomDB`: `get_*`, `query_for_*`, `*_exists` and `get_matching_atoms`. Calls without a `Keychain` are rejected; calls with a `Keychain` authorize through `AuthorizationManifest` and return only granted atoms.
- Add `KeySensitiveAtomDB` for the Keychain overloads.
- Update `AtomDBFactory` to wrap a basic backend in `ProtectedAtomDB` when the protection mode is `PROTECTED`.

Resolves #1277


src/tests/cpp/authorization_test.cc
```diff
@@ -91,13 +91,10 @@ shared_ptr<InMemoryDB> db_with_inheritance_link(string* link_handle) {
     return db;
 }
 
-AuthorizationManifest manifest_from_persistence(shared_ptr<AtomDB> atomdb,
-                                                const DummyPersistence& persistence) {
-    AuthorizationManifest manifest(atomdb);
+void add_persisted_documents(AuthorizationManifest& manifest, const DummyPersistence& persistence) {
     for (const auto& [public_key, entries] : persistence.documents) {
         manifest.add_document(make_document(public_key, false, entries));
```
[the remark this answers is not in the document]
marcocapozzoli: This dummy is only used after grant() with a schema. Unrestricted access is already covered by make_document(..., true, {}). No test rebuilds a manifest from grant_unrestricted, so I’m skipping this.

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -48,10 +46,10 @@ void AuthorizationManifest::add_document(
         RAISE_ERROR("Authorization manifest document cannot be null");
     }
 
-    auto [it, inserted] = this->profiles.emplace(
-        document->get_access_key(), AuthorizationProfile::from_document(this->atomdb, document));
-    if (!inserted) {
-        RAISE_ERROR(string("Duplicate access_key in authorization manifest: ") +
-                    document->get_access_key());
+    lock_guard<mutex> lock(this->profiles_mutex);
+    if (this->profiles.find(document->get_access_key()) != this->profiles.end()) {
```
andre-senna, line 50: We are silently rejecting updates. Is this the expected behavior?

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -48,10 +46,10 @@ void AuthorizationManifest::add_document(
...
+        return;
     }
+    this->profiles.emplace(document->get_access_key(),
+                           AuthorizationProfile::from_document(this->atomdb, document));
```
andre-senna, line 54: Do we really need to pass the entire atomdb? Does from_document() use anything other than the atomdb id?

src/atomdb/auth/AuthorizationManifest.h
```diff
@@ -39,21 +40,23 @@ class AuthorizationManifest {
...
+     * If access_key is already registered, this is a no-op. Otherwise builds
+     * an AuthorizationProfile from the document and stores it keyed by
+     * access_key. The registered-check and the insertion are atomic.
      */
     void add_document(const shared_ptr<atomdb_api_types::AccessPermissionDocument>& document);
 
    private:
     shared_ptr<AtomDB> atomdb;
     map<string, shared_ptr<AuthorizationProfile>> profiles;
+    mutable mutex profiles_mutex;
```
andre-senna, line 59: Why are you using `mutable` here?

src/atomdb/AtomDBKeySensitive.h
```diff
@@ -0,0 +1,103 @@
...
+
+using namespace std;
+using namespace atoms;
+
+namespace atomdb {
+
+/**
+ * @brief AtomDB operations that take a Keychain.
+ */
+class AtomDBKeySensitive {
```
andre-senna, line 26: KeySensitiveAtomDB sounds more natural in English

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,138 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
 // Public methods
 
 shared_ptr<Atom> ProtectedAtomDB::get_atom(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::get_atom(handle, public_key) is not implemented yet");
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
+    if (public_key.empty() || !this->ensure_registered(public_key)) {
```
andre-senna, line 33: I don't understand why we are doing lazy instantiation of the manifest here. We are not supposed to update the manifest during the lifetime of the AtomDB instance.

src/atomdb/ProtectedAtomDB.h
```diff
@@ -167,6 +165,46 @@ class ProtectedAtomDB : public AtomDB {
...
     [[noreturn]] static void raise_public_key_required(const string& method_name);
+
+    /**
+     * @brief Whether public_key may READ the atom identified by handle.
+     *
+     * This overload resolves the atom from the backend before checking
+     * permissions. Prefer the Atom overload when the atom is already
+     * available.
+     */
+    bool can_read(const string& public_key, const string& handle);
```
andre-senna, line 176: Make it inline and bring the code to the .h

src/atomdb/ProtectedAtomDB.h
```diff
@@ -167,6 +165,46 @@ class ProtectedAtomDB : public AtomDB {
...
+
+    /**
+     * @brief Whether public_key may READ atom.
+     *
+     * Avoids an additional backend lookup when the atom is already available.
+     *
+     * @return false if atom is null or the associated profile denies READ.
+     */
+    bool can_read(const string& public_key, const shared_ptr<Atom>& atom);
```
andre-senna, line 185: make it inline and bring the code to the .h

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,138 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
...
+    auto atom = this->backend->get_atom(handle);
+    if (!this->can_read(public_key, atom)) {
+        return nullptr;
+    }
+    return atom;
 }
 
 shared_ptr<Node> ProtectedAtomDB::get_node(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::get_node(handle, public_key) is not implemented yet");
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
```
andre-senna, line 45: Don't replicate the protected logic here. Just call get_atom(handle, keychain) and check if the returned atom is a actually a node before returning it.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,138 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
...
+    auto node = this->backend->get_node(handle);
+    if (!this->can_read(public_key, node)) {
+        return nullptr;
+    }
+    return node;
 }
 
 shared_ptr<Link> ProtectedAtomDB::get_link(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::get_link(handle, public_key) is not implemented yet");
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
```
andre-senna, line 58: Don't replicate the protected logic here. Just call get_atom(handle, keychain) and check if the returned atom is a actually a link before returning it.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,138 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
...
 }
 
 bool ProtectedAtomDB::atom_exists(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::atom_exists(handle, public_key) is not implemented yet");
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
+    if (public_key.empty() || !this->ensure_registered(public_key) ||
+        !this->can_read(public_key, handle)) {
+        return false;
+    }
+    return this->backend->atom_exists(handle);
```
andre-senna, line 121: Make this check first as it is cheaper. If the atom doesn't exist, return false right away. If it exists, then you check for access permission.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,138 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
...
 }
 
 bool ProtectedAtomDB::node_exists(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::node_exists(handle, public_key) is not implemented yet");
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
```
andre-senna, line 125: Don't replicate the logic of atom_exist().

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,138 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
...
+    if (public_key.empty() || !this->ensure_registered(public_key) ||
+        !this->can_read(public_key, handle)) {
+        return false;
+    }
+    return this->backend->node_exists(handle);
 }
 
 bool ProtectedAtomDB::link_exists(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::link_exists(handle, public_key) is not implemented yet");
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
```
andre-senna, line 134: Don't replicate the logic of atom_exist().

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -299,3 +382,74 @@ void ProtectedAtomDB::raise_public_key_required(const string& method_name) {
...
+}
+
+bool ProtectedAtomDB::can_read(const string& public_key, const shared_ptr<Atom>& atom) {
+    if (atom == nullptr) {
+        return false;
+    }
+    return this->manifest->is_granted(public_key, atom, AuthorizationOperation::READ);
+}
+
+bool ProtectedAtomDB::ensure_registered(const string& public_key) {
```
andre-senna, line 397: I don't understand what this method is actually doing.

andre-senna (changes requested):

src/atomdb/auth/AuthorizationManifest.cc
marcocapozzoli: The `AuthorizationProfile` does not use AtomDB directly, but it needs it to pass it to the `AuthorizationSchema`, which uses the AtomDB as a decoder in `LinkSchema::match`.

src/atomdb/auth/AuthorizationManifest.h
marcocapozzoli: It's because the `is_registered()` is `const`. But looking closely, the production caller of `AuthorizationManifest` is `ProtectedAtomDB`, which uses it as a `shared_ptr<AuthorizationManifest>`. So, I think we could remove that `const` from `is_registered` and, consequently, remove that `mutable`. What do you think?

andre-senna: If there's no reason to have the const in is_registered() body we should get rid of it and of the mutable as well.

src/atomdb/auth/AuthorizationManifest.cc
andre-senna: OK

src/atomdb/ProtectedAtomDB.cc
marcocapozzoli: This is intentional. We load a public_key into the manifest on first use `ensure_registered`, then this AtomDB instance never refreshes it. The Keychain is unknown at construction, and the backend only looks up by `public_key`.

src/atomdb/auth/AuthorizationManifest.h
marcocapozzoli: Ok. I'll remove `const` and `mutable`

src/atomdb/ProtectedAtomDB.h
marcocapozzoli: sure

src/atomdb/ProtectedAtomDB.h
marcocapozzoli: Sure, the implementation is trivial. I should have made it inline from the start

src/atomdb/ProtectedAtomDB.cc
marcocapozzoli: You're right, this is much cleaner and simpler

src/atomdb/ProtectedAtomDB.cc
marcocapozzoli: The intention here is to ensure that the `public_key` is registered in the manifest, loading its permissions if necessary. I just realized I forgot the early-return check for already registered keys. The correct implementation should be:

```
bool ProtectedAtomDB::ensure_registered(const string& public_key) {
    if (this->manifest->is_registered(public_key)) {
        return true;
    }

    // current code
}
```

src/atomdb/auth/AuthorizationManifest.cc
marcocapozzoli: Yes. Keys are not refreshed after the initial load. The skip in add_document was incorrect and has been removed.

src/atomdb/KeySensitiveAtomDB.h
```diff
@@ -0,0 +1,103 @@
...
+
+using namespace std;
+using namespace atoms;
+
+namespace atomdb {
+
+/**
+ * @brief AtomDB operations that take a Keychain.
+ */
+class KeySensitiveAtomDB {
```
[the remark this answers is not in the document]
marcocapozzoli: All in-repo C++ and Bazel references were renamed. Nothing still uses AtomDBKeySensitive or //atomdb:atomdb_key_sensitive.

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -42,16 +44,26 @@ bool AuthorizationManifest::is_granted(const string& public_key,
...
-        document->get_access_key(), AuthorizationProfile::from_document(this->atomdb, document));
+    auto access_document = this->atomdb->get_access_permissions(public_key);
+
+    if (access_document == nullptr) {
+        return false;
+    }
+
+    string access_key = access_document->get_access_key();
+    auto profile = AuthorizationProfile::from_document(this->atomdb, access_document);
+    bool inserted = this->profiles.emplace(access_key, profile).second;
```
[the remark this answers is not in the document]
marcocapozzoli: RedisMongoDB::get_access_permissions already rejects a mismatched access_key. I can add the same guard here as defense in depth. What do you think?

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -42,16 +44,26 @@ bool AuthorizationManifest::is_granted(const string& public_key,
...
+bool AuthorizationManifest::ensure_authorized(const string& public_key) {
+    lock_guard<mutex> lock(this->profiles_mutex);
+
+    if (this->profiles.find(public_key) != this->profiles.end()) {
+        return true;
     }
 
-    auto [it, inserted] = this->profiles.emplace(
-        document->get_access_key(), AuthorizationProfile::from_document(this->atomdb, document));
+    auto access_document = this->atomdb->get_access_permissions(public_key);
```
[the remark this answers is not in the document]
marcocapozzoli: Solve in 467e42a24756d73867c7fc6863689679e28bb0b7

marcocapozzoli:
@coderabbitai resolve

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -12,46 +12,78 @@ using namespace atomdb;
...
                                        shared_ptr<Atom> atom,
                                        AuthorizationOperation operation) {
-    if (!this->is_registered(public_key)) return false;
-
-    auto it = this->profiles.find(public_key);
-    if (it == this->profiles.end() || it->second == nullptr) return false;
-
-    return it->second->is_granted(atom, operation);
+    if (atom == nullptr) {
+        RAISE_ERROR("AuthorizationManifest::is_granted() requires a non-null atom");
```
andre-senna, line 28:
```suggestion
        RAISE_ERROR("AuthorizationManifest::is_granted() requires a non-null atom or a valid handle");
```

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -12,46 +12,78 @@ using namespace atomdb;
...
 bool AuthorizationManifest::is_granted(const string& public_key,
                                        const string& handle,
                                        AuthorizationOperation operation) {
-    if (!this->is_registered(public_key)) return false;
+    return this->is_granted(public_key, operation, nullptr, handle);
+}
 
-    auto it = this->profiles.find(public_key);
-    if (it == this->profiles.end() || it->second == nullptr) return false;
+bool AuthorizationManifest::ensure_profile_loaded(const string& public_key) {
```
andre-senna, line 39: This method should have a `lock_guard<mutex> lock(this->profiles_mutex);` to keep all the method synchronized. I don't see the reason to have two separate synchronized blocks. Actually, the way it's currently implemented, these two blocks allow race conditions to happen easily. If two threads pass through the first synchronization block (each one at a time), both of them will end up inserting the key and one of them will raise the error at the end.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
-void ProtectedAtomDB::raise_public_key_required(const string& method_name) {
+void ProtectedAtomDB::raise_keychain_required(const string& method_name) {
     RAISE_ERROR("ProtectedAtomDB::" + method_name +
                 "() is unavailable in protected AtomDBs. Use the public API in ProtectedAtomDB passing "
                 "a Keychain.");
 }
+
+optional<string> ProtectedAtomDB::try_get_public_key(const shared_ptr<Keychain>& keychain) {
+    string public_key = keychain ? keychain->get_public_key(this->uid_) : "";
+    if (!this->manifest->ensure_profile_loaded(public_key)) return nullopt;
```
andre-senna, line 326:
```suggestion
    if ((public_key != "")) || !this->manifest->ensure_profile_loaded(public_key)) return nullopt;
```

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
+    }
+    return authorized_handles;
+}
+
+vector<shared_ptr<Atom>> ProtectedAtomDB::filter_atoms(const vector<shared_ptr<Atom>>& original_atoms,
+                                                       const string& public_key) {
+    vector<shared_ptr<Atom>> authorized_atoms;
+    for (const auto& atom : original_atoms) {
+        if (!this->can_read(public_key, atom)) continue;
+        authorized_atoms.push_back(atom);
```
andre-senna, line 412:
```suggestion
        if (this->can_read(public_key, atom)) {
            authorized_atoms.push_back(atom);
        }
```

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
+
+    return authorized_handle_set;
+}
+
+set<string> ProtectedAtomDB::filter_handles(const set<string>& original_handles,
+                                            const string& public_key) {
+    set<string> authorized_handles;
+    for (const auto& handle : original_handles) {
+        if (!this->can_read(public_key, handle)) continue;
+        authorized_handles.insert(handle);
```
andre-senna, line 402:
```suggestion
        if (this->can_read(public_key, handle)) {
            authorized_handles.insert(handle);
        }
```

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
+    while (true) {
+        char* handle_cstr = it->next();
+        if (!handle_cstr) break;
+        string handle(handle_cstr);
+
+        if (!this->can_read(*public_key, handle)) continue;
+
+        authorized_handle_set->add_handle(handle,
+                                          original_handle_set->get_metta_expressions_by_handle(handle),
+                                          original_handle_set->get_assignments_by_handle(handle));
```
andre-senna, line 384:
```suggestion
        if (this->can_read(*public_key, handle)) {
            authorized_handle_set->add_handle(handle,
                                              original_handle_set->get_metta_expressions_by_handle(handle),
                                              original_handle_set->get_assignments_by_handle(handle));
        }
```

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
+    }
+
+#if LOG_LEVEL >= DEBUG_LEVEL
```
andre-senna, line 387: You don't need to add this #if
LOG_DEBUG() will be expanded only if LOG_LEVEL is DEBUG_LEVEL.

The point of using `#if LOG_LEVEL >= DEBUG_LEVEL`  is when you have code other then LOG_DEBUG() which you want to make sure will not be executed unless LOG_LEVEL is DEBUG_LEVEL.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
+    LOG_DEBUG("[ ProtectedAtomDB::filter_handle_set() - original handle_set: ]" +
+              std::to_string(original_handle_set->size()));
+    LOG_DEBUG("[ ProtectedAtomDB::filter_handle_set() - authorized handle_set: ]" +
+              std::to_string(authorized_handle_set->size()));
+#endif
+
+    return authorized_handle_set;
+}
+
+set<string> ProtectedAtomDB::filter_handles(const set<string>& original_handles,
```
andre-senna: filter_handles() should expect an additional parameter: `set<string>& authorized_handles` to avoid making an extra (unnecessary) copy when this function returns.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -180,122 +207,209 @@ atomdb_api_types::ProtectionMode ProtectedAtomDB::get_protection_mode() const {
...
+                                            const string& public_key) {
+    set<string> authorized_handles;
+    for (const auto& handle : original_handles) {
+        if (!this->can_read(public_key, handle)) continue;
+        authorized_handles.insert(handle);
+    }
+    return authorized_handles;
+}
+
+vector<shared_ptr<Atom>> ProtectedAtomDB::filter_atoms(const vector<shared_ptr<Atom>>& original_atoms,
```
andre-senna: filter_atoms() should expect an additional parameter: `vector<shared_ptr<Atom>>& authorized_atoms` to avoid making an extra (unnecessary) copy when this function returns.

src/atomdb/ProtectedAtomDB.cc
```diff
@@ -24,60 +29,82 @@ ProtectedAtomDB::ProtectedAtomDB(shared_ptr<AtomDB> backend) : backend(backend)
...
-    RAISE_ERROR("ProtectedAtomDB::query_for_incoming_set(handle, public_key) is not implemented yet");
+    return this->filter_handle_set(this->authorize_read(keychain, handle), [this, &handle] {
+        return this->backend->query_for_incoming_set(handle);
+    });
 }
 
 bool ProtectedAtomDB::atom_exists(const string& handle, shared_ptr<Keychain> keychain) {
-    RAISE_ERROR("ProtectedAtomDB::atom_exists(handle, public_key) is not implemented yet");
+    return this->check_handle(
+        keychain, handle, [this](const string& handle) { return this->backend->atom_exists(handle); });
```
andre-senna, line 79: You are using this pattern of passing a lambda function to the method which will actually check the public key. I don't understand the reason why you are doing it instead of just directly making the call. For example, in this case why not using something like:

```
return this->check_handle(keychain, handle, this->backend->atom_exists(handle));
```

or (even more simple):

```
return this->backend->atom_exists(handle) ? this->check_handle(keychain, handle) : false;
```

I'll not replicate this comment but I'm suggesting the revision of all the methods using lambda functions in a similar way. For some the second suggestion above may not apply, of course, but I understand that the first one does.

andre-senna (changes requested):

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -12,46 +12,78 @@ using namespace atomdb;
...
-    auto [it, inserted] = this->profiles.emplace(
-        document->get_access_key(), AuthorizationProfile::from_document(this->atomdb, document));
+    auto profile = AuthorizationProfile::from_document(this->atomdb, access_document);
+
+    lock_guard<mutex> lock(this->profiles_mutex);
+    bool inserted = this->profiles.emplace(access_key, profile).second;
     if (!inserted) {
-        RAISE_ERROR(string("Duplicate access_key in authorization manifest: ") +
-                    document->get_access_key());
+        RAISE_ERROR(string("Duplicate access_key in authorization manifest: ") + access_key);
```
ccgsnet, line 64: If we have 2 concurrent calls, the second one will RAISE here...do we really want that? Maybe we should just WARN here.

andre-senna: Or we should move the lock guard to the beginning of the method and just check if the key already exists before inserting. Just ignoring if it exists. I also don't see why we need to raise in this situation. I think it's quite possible that two threads (i.e. two different queries from the same user which arrived "together" in the query agent) end up calling this method concurrently.

andre-senna: @marcocapozzoli  Let me use this snippet as example to give a more general tip. `lockguard` behavior is such that it makes more sense when enclosed in a block. The way you used it here and in other parts of your code is misleading and tricky. It locks at the point you put it and unlocks when the function return.

The way it was designed is meant to be used at the beginning of the function so it keep the lock locked "during the function lifetime", which is clear and easy to grasp at the first glance. When you need to block only a bunch of lines inside the method (which happens frequently but is not the case in this function) you shaw either use explicit lock() unlock() (my preference, although I agree this is more style preference than rule) or lock guard INSIDE A CODE BLOCK marked with `{ }`.

ccgsnet: Agree, moving the lock to the beginning solves it.

andre-senna: Yes but you still need to do the check BEFORE trying to insert and ignore duplicates.

src/atomdb/ProtectedAtomDB.cc
marcocapozzoli: I didn't understand why `public_key != ""`. Technically, that condition will always be true. I think you meant `public_key == ""`. That validation is already handled inside `ensure_profile_loaded()`.

andre-senna: Actually I meant `public_key == `. Making the check here makes more sense because it'd avoid calling the method `ensure_profile_loaded()`.

src/atomdb/ProtectedAtomDB.cc
marcocapozzoli: I originally implemented it this way because it was the first approach that came to mind, and since it worked, I moved on. But your suggestion is simpler and cleaner, so I'll adopt it.

marcocapozzoli: Regarding the method you mentioned (`atom_exists`), I agree that your approach is simpler and cleaner than passing a function. I'll update the three methods (`atom_exists`, `node_exists`, and `link_exists`) to use the ternary operator.

I originally used the lambda approach to keep it consistent with five other methods that follow the same pattern: `query_for_pattern`, `query_for_incoming_set`, `atoms_exist`, `nodes_exist`, and `links_exist`.

For those methods, I use a lambda on purpose because I don't want to run an expensive database operation (for example, `query_for_pattern`) before checking the caller's key. If the key is missing or invalid, we can return early and avoid the database call entirely.

for example: If you look at `check_handles()`, it follows the same idea: it checks the key first and only then runs the function.

```
set<string> ProtectedAtomDB::check_handles(shared_ptr<Keychain> keychain,
                                           const function<set<string>()>& exists_many) {
    auto public_key = this->try_get_public_key(keychain);
    if (!public_key) return {};
    return this->filter_handles(exists_many(), *public_key);
}
```

ccgsnet (approved):

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -12,46 +12,73 @@ using namespace atomdb;
...
 
-    return it->second->is_granted(atom, operation);
+    auto access_document = this->atomdb->get_access_permissions(public_key);
+    if (access_document == nullptr) {
+        return false;
+    }
+
+    string access_key = access_document->get_access_key();
+    if (access_key != public_key) {
+        return false;
```
andre-senna:
```suggestion
        LOG_ERROR("Corrupted or misplaced access document for key: " + public_key);
        return false;
```

src/atomdb/auth/AuthorizationManifest.cc
```diff
@@ -12,46 +12,73 @@ using namespace atomdb;
...
+                                       AuthorizationOperation operation,
+                                       shared_ptr<Atom> atom,
+                                       const string& handle) {
+    shared_ptr<AuthorizationProfile> profile;
+    {
+        lock_guard<mutex> lock(this->profiles_mutex);
+        auto it = this->profiles.find(public_key);
+        if (it == this->profiles.end() || it->second == nullptr) return false;
+        profile = it->second;
     }
```
andre-senna:
```suggestion
    lock_guard<mutex> lock(this->profiles_mutex);
    
    shared_ptr<AuthorizationProfile> profile;
    auto it = this->profiles.find(public_key);
    if (it == this->profiles.end() || it->second == nullptr) return false;
    profile = it->second;
```

andre-senna (approved):

